PR #13, opened 2018-03-09
Adds label printing option through Zebra Print Connect

bellerbrock:


bellerbrock:
Hey @trife, any more thoughts on the barcode printing functionality? A few sweetpotato programs have been using the forked version that includes the label printing option, and some additional programs have asked me how they can start using it. 

I'm reluctant to tell them because I don't want to create duplication, and I don't think my implementation is very good. @chaneylc 's printing option in Intercross is cleaner. And you're idea of implementing it as a trait is a good one. 

I can update my fork and pull request with those changes if you like, but I thought I ask first since I realize you or Chaney could probably do a better job and may prefer to do it yourselves.

trife:
@bellerbock, I'm definitely still working through this. I've had to prioritize Chaney onto other apps but he has developed an even better solution for printing that I should be able to integrate. I think it's prudent to first integrate Selby's fork that makes adding new traits significantly easier. I'll work on that and then may have you redo your pull request to fit those guidelines if that works for you.

trife:
Added as new trait.

PR #33, opened 2019-10-09
Check for duplicate ids when importing brapi fields

ctucker3:
Currently with csv imports, Field Book does not allow duplicate id's in the unique_id column in the plot table. This feature adds a check when importing a BrAPI field to make sure we are not importing duplicate unique plot ids.


ctucker3:
[FIE-56 - Check duplicate unique_id's from importing from brapi](https://trello.com/c/ckP6hr05/38-fie-56-check-duplicate-uniqueids-from-importing-from-brapi)

PR #34, opened 2019-10-10
Fie 60

ctucker3:
This fixes an issue on the Brapi Export page. When the export button was clicked, and a 401 response was returned, the login process that began on the export page after that did not successfully get the deep link response. 

Another addition in the branch besides the auth loop fix was to reformat the brapi login code in the preferences menu. Reformatting consisted of using native android functions for the activity flow, instead of the hack arounds that were in place before. Also, a general cleaning up of button visibility and text code.

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/preferences/PreferencesActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/preferences/PreferencesActivity.java
@@ -85,9 +85,6 @@ public class PreferencesActivity extends AppCompatActivity {
             getSupportActionBar().setHomeButtonEnabled(true);
         }
 
-        // Check if our activity was started up with brapi auth deep link.
-        brapiControllerResponse = BrAPIService.checkBrapiAuth(this);
-
         // This is not related to the deep link, load normally.
         preferencesFragment = new PreferencesFragment();
         getFragmentManager().beginTransaction()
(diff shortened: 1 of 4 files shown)


ctucker3:
[FIE-60 - Auth loop if authentication is entered after export button click](https://trello.com/c/7ba24naY/42-fie-60-auth-loop-if-authentication-is-entered-after-export-button-click)

app/src/main/java/com/fieldbook/tracker/preferences/PreferencesActivity.java
```diff
@@ -115,21 +113,13 @@ public void onResume() {
         super.onResume();
 
         // If our preference page was resumed, we will want to see if it was resumed from a deep link.
-        if (brapiControllerResponse.status == null) {
```
nickpalladino: Discussed change from having call in OnCreate to only using OnResume which is why the null check is no longer there

app/src/main/java/com/fieldbook/tracker/preferences/PreferencesFragment.java
```diff
@@ -135,18 +129,45 @@ public boolean onPreferenceClick(Preference preference) {
...
+
+    public void setButtonView() {
+
+        String brapiToken = prefMgr.getSharedPreferences().getString(PreferencesActivity.BRAPI_TOKEN, null);
+        String brapiHost = prefMgr.getSharedPreferences().getString(BRAPI_BASE_URL, null);
+
+        if(brapiHost != null && !brapiHost.equals(getString(R.string.brapi_base_url_default))) {
+
+            if (brapiToken != null) {
+                // Show our reauthorize button and remove logout button
```
nickpalladino: comment part about removing logout button is not accurate

PR #36, opened 2019-10-10
Fie 58

ctucker3:
Modifies the brapi trait page so that when the traits are paginated, if selected, the selections will persist between pages. 

Also fixes a bug with the pagination, where the prev and next buttons will change the value of the current page, even though they are not executing the brapi call. The buttons now do nothing when clicked if there is only one page of traits. 

Also increased the number of traits per page to 15.

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/brapi/BrapiTraitActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/BrapiTraitActivity.java
@@ -10,6 +10,7 @@ import android.view.MenuItem;
 import android.view.View;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
+import android.widget.Button;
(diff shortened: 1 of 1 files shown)


ctucker3:
[FIE-58 - Trait pagination. Keep selected traits when scrolling to next page](https://trello.com/c/XZcrbjvD/40-fie-58-trait-pagination-keep-selected-traits-when-scrolling-to-next-page)

ctucker3:
Good thinking! We can just set them to disabled if there is only one page. I'll add it tomorrow

ctucker3:
Button behavior has been altered to hide the buttons to match the options available to the user. For example, if the user is on page 1 out of 2, the previous page button will be invisible. If the user is on page 2 of 2, the next button will be invisible but the previous button will be visible.

PR #37, opened 2019-10-11
Brapi synced trait delete NA handling

nickpalladino:
Moved from onFocusChangeListener to managing clearing hints in the appropriate trait layouts as it wasn't working for all cases.

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/MainActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/MainActivity.java
@@ -296,14 +296,6 @@ public class MainActivity extends AppCompatActivity {
             }
         });
 
-        etCurVal.setOnFocusChangeListener(new View.OnFocusChangeListener() {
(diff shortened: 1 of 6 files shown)


nickpalladino:
[FIE-55 - Don't allow empty values to be saved in brapi synced trait values, instead show that NA will be saved](https://trello.com/c/XpN9nuk0/37-fie-55-dont-allow-empty-values-to-be-saved-in-brapi-synced-trait-values-instead-show-that-na-will-be-saved)

ctucker3:
Looks good. 

After discussion with Nick, barcode was left out because that data type doesn't work anyway. A few specific data type classes (but not all) have explicit setting of the hint to blank on trait load layout because those traits don't have on focus change events. 

When a person deletes a brapi synced trait, the NA value is set as the value, when the person backspaces a trait, the NA text hint is set. Both instances described write to the database.

PR #39, opened 2019-10-14
Excluded extraneous columns in trait csv export

nickpalladino:
Excluded csv trait export columns that were added to the traits table from our brapi work. This keeps the format the same as it was before for csv trait trt files.


nickpalladino:
Ya, agree that would be less error prone. I can do that.

PR #54, opened 2020-01-24
Inamori branch

m-inamori:
I divided FieldFile into FieldFileCSV and FieldFileExcel in order to absorb the difference between CSV and Excel.
I separated NewTraitDialog from TraitEditorActivity.
I packed trait layouts, the part of trait changing elements, and the part of range changing elements.

In the process, I found some bugs and fixed them:
1. In FieldEditorActivity etc. we make it to the background and leave it a few minutes (2 minutes in my environment), and make it to the foreground. Then Field Book crashes.
2. When editing a trait in FieldEditorActivity not in English, the previous format is kept.
3. In FieldEditorActivity, relationship among initial, minimum, and maximum values is not checked.
4. MultiCategory trait does not work.
5. In Category and MultiCategory, category duplication is not checked.
6. In MainActivity, if the minimum or maximum value is set for the trait for which data is being entered, the data will not be saved.
7. In MainActivity, when long pressing leftRange, Field Book crashes in HIDE_ENTRIES_WITH_DATA mode.

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/ConfigActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/ConfigActivity.java
@@ -160,7 +160,6 @@ public class ConfigActivity extends AppCompatActivity {
 
         invalidateOptionsMenu();
         loadScreen();
-
     }
 
     @Override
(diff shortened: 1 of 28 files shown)


m-inamori:
I restored moveToSearchs.
But, sorry, I forgot to restore one of them ( Line 569).
            int[] rangeID = rangeBox.getRangeID();
            if (rangeID != null) {
                moveToSearch("search",rangeID, searchRange, searchPlot, null);
            }

PR #97, opened 2020-03-12
Inamori branch

m-inamori:
# Description

I made it possible to enter trait values with bar-code.
In Iwata Lab, we use bar-code measures.
A bar-code is attached to the measures every 1cm.
We read the bar-code with a bar-code reader and use it as a measured value.
It would be very useful if we could read with a smartphone instead of a barcode reader.

I added buttons to launch the bar-code reader app in three places.
These are NumericalTraitLayout, TextTraitLayout and CounterTraitLayout.

## Type of change

# How Has This Been Tested?

I tested the operation of entering Plot ID and values on a real machine.
machine: ASUS_X00HD,
Android: 7.1.1

- [ ] Test A
- [ ] Test B

**Test Configuration**:
* Hardware:
* SDK:

# Checklist:


trife:
Hi @m-inamori , do you think it would make more sense to have a separate trait layout solely for barcode data? Is there value in repeating the barcode functionality on only a subset of the layouts instead of it having its own layout?

m-inamori:
Hi, @trife.
First, I also thought that a new trait layout must be ready for bar-code system.
But, each trait layout is corresponding to a type of trait objects.
It may be better to add a bar-code option to the trait editor, and if it is checked another trait layout comes up.

trife:
Hi @m-inamori , after discussing, I think it makes more sense right now to have a separate layout specifically for barcodes. I'll work on this and hopefully get it added for v4.3. We may revisit your approach in the future if a separate layout is insufficient.

Issue #129, opened 2020-04-30
[BUG] Only 1000 fields (studies) are returned from BrAPI


nahuel-soldevilla:
**Describe the bug**
I'm not able to retrieve more than a 1000 results importing fields using the BrAPI workflow. 

**To Reproduce**

1. Setup BrAPI endpoint in Setting and authorize
2. Go to Fields and import Fields using BrAPI

**Expected behavior**
I should be able to paginate through results if there is more than one page.

**Device Info (please complete the following information):**
 - Model: Android Studio emulator - Pixel 2 API 28
 - OS: Android 9
 - Field Book Version 4.3

#

I've created the issue for the studies endpoint, but it applies also to other entities (observationUnits, traits).
We are willing to contribute with this feature.

cc @trife @BrapiCoordinatorSelby @mcrimi

PR #152, opened 2020-09-01
https://github.com/PhenoApps/Field-Book/issues/129

trife:
feature request for issue #129

Still needs testing. Modified BrAPI service to take a pagination parameter which is now a preference in preferences_brapi.xml

Pagination preference must be an integer, or else it defaults to 1000.

Modified all locations of BrAPI service initialization, added buffer preference icon, added preference key BRAPI_PAGINATION.


trife:
@nahuel-soldevilla can you review this pull request and confirm that it fixes #129 ?

nahuel-soldevilla:
Hi @trife 

First of all, thank for this! and thanks @chaneylc too.

I think it certainly helps with the problem, because with this we can increase now the page size and retrieve more results, and if any of the rows we need is beyond the 1000 number, we can reach it now.

What I had in mind initially is that maybe we can add two more buttons (next/prev) and use it to navigate through pages, and as I'm writing this I now realize that this behavior is already implemented for traits loading from BrAPI:
fieldbook-trait-pagination
and uses a page size of 15. Do you think it's possible to implement the same thing for the other BrAPI activities? and combine it with this new pageSize shared pref?

cc @BrapiCoordinatorSelby

BrapiCoordinatorSelby:
@trife @nahuel-soldevilla I made the updates to this branch to allow paging everywhere. I made the "Next/Prev" buttons into a partial layout, and created BrAPIPaginationManager.java to generalize the functionality.

As it is now, it makes a new BrAPI call every time you change pages. A future update might add a page cache to minimize HTTP requests. 

Also, I changed the functionality when grabbing the plots and traits for a study. Previously, there was a limit and it would only download the first 1000 plots/traits. Now it will download ALL the plots/traits, by getting all the pages until the server doesn't have any more. I put a hard limit at 50,000 items for safety (1000 plots x 50 pages), but this can be adjusted pretty easily if needed.

nahuel-soldevilla:
Thanks @BrapiCoordinatorSelby!
I tested this on BMS and didn't find any issue so far.

PR #183, opened 2020-11-05
Database refactor

chaneylc:
# Description

The motivation of this pull request is to modularize and refactor the back-end for a new schema update. Every table in the new schema has a respective model, along with new queries and functions that replace the DataHelper functions. 

The new schema mapping: 
[field_book_database_mapping_20200902.xlsx](https://github.com/PhenoApps/Field-Book/files/5497462/field_book_database_mapping_20200902.xlsx)

The new back-end files are in com.fieldbook.tracker.database. Model files are included for each table in the new schema. Most queries belong to a given module, otherwise they're included in the DataHelper2.kt utility file. For example, 'updateExpTable()' in DataHelper is equivalent to 'StudyModel.updateExpTable()' in the StudyModel file because the old schema table exp_id is now the studies table which has a StudyModel Kotlin file implementation. But currently the ObservationUnitProperty (new range table) table has no individual model file, so it has functions in the DataHelper2.kt file.

I commented out a line in DataHelper.java line 1686, this makes the function more test-friendly, but this might cause issues in ConfigActivity if the database wasn't opened before this call.

Furthermore, this pull request contains two new files for testing the back-end schema change. One file, DatabaseBatchTest is for testing a set of databases. DatabaseTest is for testing individual databases. The tests expect an old database to be present, which is loaded at the beginning of each test, before the test is run the new schema tables are created. 

There is a test for every function in DataHelper, each test runs the old back-end and notifies if it fails. If the old back-end does not fail, the new back-end is run and the queries are compared. Query comparison is dependent on the function, order is matched when necessary. Generally, if a function returns a cursor then the column names and values are matched; otherwise, if an object s.a a FieldObject is returned, then each member within the object is checked. There are some exceptions, including last_synced_time with observations.

Some issues with the DataHelper.java back-end have been found with these tests. Specifically, some databases seem to have row data that has BLOB, and NUMERIC affinity. This causes some data to be missed in the export functions, s.a url's, and some integer values. These tests can be used to reproduce SQLite errors in regards to the 64 limit join queries. Another issue found in DataHelper was the sanitization of queries. Two distinct problems occurred when blank columns or identifiers were used, specifically when plot attribute names were blank values. Secondly, when identifiers were quoted with ' and contained '.  https://sqlite.org/lang_keywords.html brackets or grave accents are recommended for this.

Kotlin is now a dependency.

## Type of change

- [X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
I marked this as a breaking change because of the line 1686 comment in DataHelper. We might need a better way to ensure the database is open before queries are run in CollectActivity.

# How Has This Been Tested?

The app runs on emulator, but runs into database access issues in CollectActivity.


app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
```diff
@@ -0,0 +1,247 @@
...
+                .toTypedArray()
+
+        } ?: emptyArray()
+
+        /**
+         * In this case parent is the variable name and trait is the format
+         */
+        fun insertUserTraits(rid: String, parent: String, trait: String,
+                             userValue: String, person: String, location: String,
+                             notes: String, exp_id: String, observationDbId: String?,
```
timparsons, line 37: Is there any intention of storying `notes` and `observationDbId`?  If not, I'd recommend removing them

timparsons (changes requested):
Overall, looks like a really nice improvement to the structure of the code.  I do have some comments and suggestions:

- Each of the schema models has a PK and FK constant.  Why aren't the rest of the columns in the table represented as constants?

- There are a number of queries that are doing comparisons using `LIKE`, when it appears they should be using `=`.  For example, the where clause in `ObservationDAO->getObservation` (line 113).

- Other than a handful of methods, the DataHelper doesn't look like it does much other than wrap around the DAOs.  Are there plans to refactor current calls to the DataHelper, and instead call directly to the DAOs?

- Look into database versioning
	This is a large effort in and of itself, so I wouldn't suggest doing it within this effort.
	
	There are libraries that exist that can version a database, and upgrade it when needed.  Your code base would have migration scripts that start at the very beginning of the initial creation of the database, and then have additional scripts added over time for changes made (updates, creation of new tables, modification of existing data, etc), and the library handles applying these scripts to the database automatically.  Flyway (flywaydb.org) is a nice library that only requires you to make sql files with a specific name format in order for it to know the order in which to apply the patches.

- Look into dependency injection
	This is a large effort in and of itself, so I wouldn't suggest doing it within this effort.

	https://developer.android.com/training/dependency-injection

	@BrapiCoordinatorSelby and I both have extensive experience using Google Guice, and have enough knowledge of Dagger to help with any questions.  Here's a guide for using Dagger in Android: https://developer.android.com/training/dependency-injection/dagger-android

	I'm noticing how the DataHelper is created for each Activity within the `onCreate` method.  I would suggest that `DataHelper` become a singleton, and is therefore only created once.  If using dependency injection, this is very easy to define the `DataHelper` as a singleton.  Activity classes would also become more modular and more testable because they wouldn't create new instances of `DataHelper` (or any other dependencies) on their own, but rather, have it injected.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,591 @@
...
+                        migrateTo(db, migrateFromTableName, tableName, columnDefs, migratePattern)
+                    }
+
+                    db.execSQL(sVisibleObservationVariableView)
+
+                    db.setTransactionSuccessful()
+
+                } catch (e: Exception) {
+
+                    e.printStackTrace()
```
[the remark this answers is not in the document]
chaneylc: From my understanding of the docs, a rollback automatically occurs if setTransactionSuccessful is not called. 
https://developer.android.com/reference/androidx/sqlite/db/SupportSQLiteDatabase
db.beginTransaction();
   try {
     ...
     db.setTransactionSuccessful();
   } finally {
     db.endTransaction();
   }

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,591 @@
...
+         * Creates Tables for new schema.
+         * 1. deletes any old table that exists referencing new schema rows. e.g StudyModel
+         * 2. Builds and runs a create table statement based on the table class.
+         * 3. Queries the original db for exp_id rows
+         * 4. Populates table based on previous query by translating
+         *      column headers using a migration pattern mapping.
+         */
+        fun createTables(db: SQLiteDatabase, traits: ArrayList<TraitObject>) {
+
+            deleteTables()
```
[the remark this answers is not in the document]
chaneylc: Nice catch, I moved this into the transaction.

app/src/main/java/com/fieldbook/tracker/database/SqlUtil.kt
```diff
@@ -0,0 +1,214 @@
...
+        val result = function(db)
+
+        db.setTransactionSuccessful()
+
+        return@withDatabase result
+
+    } catch (e: Exception) {
+
+        e.printStackTrace()
+
```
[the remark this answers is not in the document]
chaneylc: This function is no longer used.

app/src/main/java/com/fieldbook/tracker/database/SqlUtil.kt
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+    } catch (e: IllegalArgumentException) {
+
+        e.printStackTrace()
+
+    } catch (ie: IndexOutOfBoundsException) {
+
+        ie.printStackTrace()
```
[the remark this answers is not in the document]
chaneylc: I moved these catches to the withDatabase function. The toFirst and toTable functions are called with withDatabase on the callstack, so these exceptions are now propagated to the withDatabase exception handler.

app/src/main/java/com/fieldbook/tracker/database/SqlUtil.kt
```diff
@@ -0,0 +1,214 @@
...
+
+        } while (moveToNext())
+
+    } catch (ie: IndexOutOfBoundsException) {
+        ie.printStackTrace()
+    } catch (e: IllegalArgumentException) {
+        e.printStackTrace()
+    } catch (noElem: NoSuchElementException) {
+        noElem.printStackTrace()
+    }
```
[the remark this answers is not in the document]
chaneylc: Again, I moved these to the withDatabase function.

app/src/main/java/com/fieldbook/tracker/database/SqlUtil.kt
```diff
@@ -0,0 +1,214 @@
...
+
+    }
+
+} catch (npe: NullPointerException) {
+    //npe.printStackTrace()
+    null
+} catch (e: Exception) {
+    e.printStackTrace()
+    null
+}
```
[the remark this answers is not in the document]
chaneylc: These are not propagated because there is no guarantee that the calling method is catching errors. If a backend error (SQLiteException) occurs, then the app would crash. The original idea was to return null if any error occurred. I suppose when we implement ViewModels, these errors could be handled there instead, which might be a bit cleaner.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,591 @@
...
+
+                models.forEach {
+//            println(it)
+                    try {
+                        db.insertWithOnConflict(to, null, it.toContentValues(), SQLiteDatabase.CONFLICT_IGNORE)
+                    } catch (constraint: SQLiteConstraintException) {
+                        //constraint.printStackTrace()
+                    } catch (exp: Exception) {
+                        exp.printStackTrace()
+                    }
```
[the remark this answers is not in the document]
chaneylc: I cleaned this up and moved the catches to the calling method.

app/src/main/java/com/fieldbook/tracker/database/SqlUtil.kt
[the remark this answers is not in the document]
timparsons: >I suppose when we implement ViewModels, these errors could be handled there instead, which might be a bit cleaner.

I think that would be the right direction to go.  I would think you would want the view responsible for making the call to the database to know about any errors so you can inform the end-user.

chaneylc:
Thanks for putting time into reviewing this pull request. You mention a lot of useful libraries and helpful fixes :)

>     * Each of the schema models has a PK and FK constant.  Why aren't the rest of the columns in the table represented as constants?
That part was a little rushed, you're correct they should be added as constants.
 
>     * There are a number of queries that are doing comparisons using `LIKE`, when it appears they should be using `=`.  For example, the where clause in `ObservationDAO->getObservation` (line 113).
These operators are used in the original backend queries, I didn't want to break any functionality. This is also something I'm curious about because if no patterns are used in these operations then we should switch to the equality operator.

>     * Other than a handful of methods, the DataHelper doesn't look like it does much other than wrap around the DAOs.  Are there plans to refactor current calls to the DataHelper, and instead call directly to the DAOs?
Not currently, but eventually we can implement ViewModels that will call the dao queries.

>     * Look into database versioning
>       This is a large effort in and of itself, so I wouldn't suggest doing it within this effort.
>       There are libraries that exist that can version a database, and upgrade it when needed.  Your code base would have migration scripts that start at the very beginning of the initial creation of the database, and then have additional scripts added over time for changes made (updates, creation of new tables, modification of existing data, etc), and the library handles applying these scripts to the database automatically.  Flyway (flywaydb.org) is a nice library that only requires you to make sql files with a specific name format in order for it to know the order in which to apply the patches.
Interesting, I will keep this in mind. Thanks!

>     * Look into dependency injection
>       This is a large effort in and of itself, so I wouldn't suggest doing it within this effort.
>       https://developer.android.com/training/dependency-injection
>       @BrapiCoordinatorSelby and I both have extensive experience using Google Guice, and have enough knowledge of Dagger to help with any questions.  Here's a guide for using Dagger in Android: https://developer.android.com/training/dependency-injection/dagger-android
>       I'm noticing how the DataHelper is created for each Activity within the `onCreate` method.  I would suggest that `DataHelper` become a singleton, and is therefore only created once.  If using dependency injection, this is very easy to define the `DataHelper` as a singleton.  Activity classes would also become more modular and more testable because they wouldn't create new instances of `DataHelper` (or any other dependencies) on their own, but rather, have it injected.
This is a great discussion topic for the overall architecture of FieldBook. I think eventually we can modularly implement a more jetpack-like architecture, and here Dagger Android / Hilt / ButterKnife will fit perfectly. After completing this back-end refactoring, we can focus on implementing DI for ViewModels (and other Android classes).

BrapiCoordinatorSelby:
@chaneylc 
quick note: do a `gradle :app:clean` then a `gradle :app:compileDebugKotlin` and you'll get a list of Kotlin warnings. Take a quick look through them to catch unused variables etc. I see about 14. Not a big deal, but will help clean up a little.
Stupid gradle seems to ignore them unless you explicitly clean first 🤦

app/src/main/java/com/fieldbook/tracker/activities/ConfigActivity.java
```diff
@@ -229,10 +230,11 @@ public void onItemClick(AdapterView<?> av, View arg1, int position, long arg3) {
 
                         break;
                     case 2:
```
BrapiCoordinatorSelby, line 232: Suggestion: use the same code for both `case 2` and `case 3`, perhaps move this code to a method

BrapiCoordinatorSelby:
FYI I opened #195 so one of us will remember to go look at the timestamp inconsistencies.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
```diff
@@ -0,0 +1,410 @@
...
+                            put("last_synced_time", it.lastSyncedTime.format(DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSSZ", Locale.getDefault())))
+                        },
+                        "${Observation.PK} = ?", arrayOf(it.fieldBookDbId))
+
+            }
+        }
+
+
+        //TODO
+        fun updateImage(image: Image, writeLastSyncedTime: Boolean) {
```
BrapiCoordinatorSelby, line 330: Is there a reason why this method follows a different `withDatabase` pattern? It looks equivalent, just curious why the difference

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -971,232 +1046,276 @@ public FieldObject getFieldObject(Integer exp_id) {
...
+//        return data;
     }
 
     /**
      * Get observation data that needs to be saved on edits
      */
     public Observation getObservation(String plotId, String parent) {
 
-        Observation o = new Observation();
+        Observation o = ObservationDao.Companion.getObservation(plotId, parent);
```
BrapiCoordinatorSelby, line 1186: Make this a `return` statement like others

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -971,232 +1046,276 @@ public FieldObject getFieldObject(Integer exp_id) {
...
     /**
      * Returns the primary key for all ranges
      */
     public int[] getAllRangeID() {
-        //TODO check for range table, if not exist create
-        Cursor cursor = db.query(RANGE, new String[]{"id"}, null, null,
-                null, null, "id");
 
-        int[] data = null;
+        if (!isTableExists("ObservationUnitProperty")) {
```
BrapiCoordinatorSelby, line 1266: Is this ever expected to happen? Is there a use case where it is important to get the `observationUnitProperties` for the first field? even though that field has not been selected yet and there might not even be a first field to get? Maybe this should be part of a unit test setup rather than built into the method?

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -1368,65 +1499,70 @@ public void updateTraitByValue(String rid, String parent, String value, String n
...
+//        } catch (Exception e) {
+//            return null;
+//        }
     }
 
     /**
      * Returns the column names for the range table
      */
     public String[] getRangeColumnNames() {
         if (db == null || !db.isOpen()) db = openHelper.getWritableDatabase();
```
BrapiCoordinatorSelby, line 1537: delete this line? unnecessary with the new DAO classes and `withDatabase` helper method

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -1615,267 +1780,312 @@ public long insertTraits(TraitObject t) {
...
 
-        if (ex) {
-            // get export date
-            ContentValues cv = new ContentValues();
-            cv.put("date_export", timeStamp.format(Calendar.getInstance().getTime()));
-            db.update(EXP_INDEX, cv, "exp_id=" + exp_id, null);
-        }
-
-        cursor.close();
+        ConfigActivity.dt.open();
```
BrapiCoordinatorSelby, line 1864: this line is probably unnecessary with new DAO layer

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -1615,267 +1780,312 @@ public long insertTraits(TraitObject t) {
...
-        db.execSQL("DELETE FROM " + PLOTS + " WHERE exp_id = " + exp_id);
-        db.execSQL("DELETE FROM " + PLOT_ATTRIBUTES + " WHERE exp_id = " + exp_id);
-        db.execSQL("DELETE FROM " + PLOT_VALUES + " WHERE exp_id = " + exp_id);
-        db.execSQL("DELETE FROM " + USER_TRAITS + " WHERE exp_id = " + exp_id);
-    }
 
-    public void switchField(int exp_id) {
-        Cursor cursor;
+        //TODO add optional cascade delete
+        StudyDao.Companion.deleteField(exp_id);
```
BrapiCoordinatorSelby, line 1916: This is not equivalent to what was originally there. Orphaned records might still exist after `deleteField` is called

BrapiCoordinatorSelby: never mind, just found the cascade on delete code

trife: Chaney and I talked about this a bit and ultimately decided that skipping the cascade delete and reimporting the same field as a new field with disconnected phenotypes was the safest implementation for users and would allow old phenotypes to be recovered if needed.

BrapiCoordinatorSelby: Cool cool. As long as its a conscious choice and not a typo then I'm satisfied.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,605 @@
...
+
+/**
+ * Migration helper class that contains all schema refactoring patterns, table names, and private/foreign key references.
+ *
+ * Each model class has a respective class here that contains a Schema static object.
+ * These classes should never be initialized and are just used in createTables to migrate the database version.
+ *
+ *
+ */
+class Migrator {
```
BrapiCoordinatorSelby, line 20: Consider renaming this class (or the methods inside it, or at least a comment at the top) to specifically reference Field Book DBv9. 
It looks like this class only handles the upgrade from DBv8 to DBv9 and is not prepared to handle the any future updates. So it is not a generic `Migrator` but a specific `DBv9Migrator`. You could separate the out generic methods into a parent class, but its probably not worth the effort right now. And, as @timparsons mentioned, there are DB upgrade libraries to handle this sort of thing in the future.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,605 @@
...
+    class ObservationUnit private constructor() {
+
+        companion object Schema {
+            const val PK = "internal_id_observation_unit"
+            const val FK = "observation_unit_id"
+            const val migrateFromTableName = "plots"
+            const val tableName = "observation_units"
+            val columnDefs by lazy {
+                mapOf(PK to "INTEGER PRIMARY KEY AUTOINCREMENT",
+                        "study_db_id" to "INT REFERENCES ${Study.tableName}(${Study.PK}) ON DELETE CASCADE",
```
BrapiCoordinatorSelby, line 361: `study_db_id` here is not actually a studyDbId, it is a studyPK (aka exp_id). 
I think the migration will be ok, but if you load a new study and try to put a BrAPI studyDbId in `observationUnit.study_db_id` the foreign key constraint will break.

BrapiCoordinatorSelby: reading further, `study_db_id` here should be `Study.FK`. 
I still think the value of `Study.FK`should change, but at least it will be consistent across all tables

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,605 @@
...
+
+        /**
+         * Creates Tables for new schema.
+         * 1. deletes any old table that exists referencing new schema rows. e.g StudyModel
+         * 2. Builds and runs a create table statement based on the table class.
+         * 3. Queries the original db for exp_id rows
+         * 4. Populates table based on previous query by translating
+         *      column headers using a migration pattern mapping.
+         */
+        fun createTables(db: SQLiteDatabase, traits: ArrayList<TraitObject>) {
```
BrapiCoordinatorSelby, line 111: Why are traits passed in as an argument?
`DataHelper.getAllTraitObjects()` only exists as part of the migrate process and should probably be moved out of DataHelper into Migrator.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,605 @@
...
+                        ))
+
+//                            println("$rowid Inserting ${attrValue.key} = ${attrValue.value} at ${attrIds[attrValue.key]}")
+//                        }
+                    }
+                }
+
+                with(Observation) {
+                    migrateTo(db, migrateFromTableName, tableName, columnDefs, migratePattern)
+                }
```
BrapiCoordinatorSelby, line 190: why is this `migrateTo` hidden down here?

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -0,0 +1,605 @@
...
+        companion object Schema {
+            const val PK = "internal_id_observation"
+            const val FK = "observation_id"
+            const val migrateFromTableName = "user_traits"
+            const val tableName = "observations"
+            val columnDefs by lazy {
+                mapOf(PK to "INTEGER PRIMARY KEY AUTOINCREMENT",
+                        ObservationUnit.FK to "TEXT",
+                        Study.FK to "INT REFERENCES ${Study.tableName}(${Study.PK}) ON DELETE SET NULL",
+                        ObservationVariable.FK to "INT",
```
BrapiCoordinatorSelby, line 398: missing foreign key constraints?

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
```diff
@@ -0,0 +1,410 @@
...
+
+        } ?: 0
+
+        /**
+         * Unit -> plot/unit internal id
+         * variable name -> trait/variable readable name
+         * variable -> trait/variable db id
+         * format -> trait/variable observation_field_book_format
+         */
+        fun randomObservation(unit: String, uniqueName: String, format: String, variableName: String, variable: String): ObservationModel = ObservationModel(mapOf(
```
BrapiCoordinatorSelby, line 369: I assume this is for testing. Delete or move to a JUnit class as test prep

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitDao.kt
```diff
@@ -0,0 +1,48 @@
...
+
+            arrayOf(*db.query(ObservationUnit.tableName,
+                    where = "${Study.FK} = ?",
+                    whereArgs = arrayOf(eid.toString())).toTable()
+                    .map { ObservationUnitModel(it) }
+                    .toTypedArray())
+
+        } ?: emptyArray()
+
+        fun getAll(): Array<ObservationUnitModel> = withDatabase { db ->
```
BrapiCoordinatorSelby, line 40: I assume this is for testing. Delete or move to a JUnit class as test prep

app/src/main/java/com/fieldbook/tracker/activities/ConfigActivity.java
chaneylc: Sounds good, I'll include this fix in the next commit.

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
chaneylc: You're right, this shouldn't ever happen, I'll clean this up. Thanks for the catch :)

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
chaneylc: This one will need a bit more testing. I think we should modify the activity code to always ensure the db is open before making dao calls.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
chaneylc: Good catch, do you have a suggestion for what Study.FK should be?

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
chaneylc: The idea was to migrate all observation dependencies s.a observation variables and observation units first. I believe we would get a foreign key constraint violation otherwise.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
chaneylc: This one is tricky because its specific to a feature in FieldBook. We actually keep deleted observations, in case the user accidentally deleted their work. That being said, we could probably create a constraint and have it set to null on delete, similar to the study fk.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
chaneylc: No reason here :facepalm:  Thanks for the reminder I meant to clean this function up.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt
```diff
@@ -0,0 +1,216 @@
...
+import com.fieldbook.tracker.database.toFirst
+import com.fieldbook.tracker.database.toTable
+import com.fieldbook.tracker.database.withDatabase
+import com.fieldbook.tracker.objects.RangeObject
+
+class ObservationUnitPropertyDao {
+
+    companion object {
+
+        internal fun selectAllRange(): Array<Map<String, Any?>>? = withDatabase { db ->
```
BrapiCoordinatorSelby, line 17: unused method

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt
```diff
@@ -0,0 +1,216 @@
...
+            val table = db.query(sObservationUnitPropertyViewName,
+                    select = arrayOf("id"),
+                    orderBy = "id").toTable()
+
+            table.map { (it["id"] as java.lang.Integer) }
+                    .toTypedArray()
+        } ?: emptyArray()
+
+        //TODO original code uses switchField if object is null
+        fun getRangeByIdAndPlot(firstName: String,
```
BrapiCoordinatorSelby, line 31: unused method

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableAttributeDao.kt
```diff
@@ -0,0 +1,35 @@
...
+
+        fun getAttributeIdByName(name: String): Int = withDatabase { db ->
+
+            db.query(ObservationVariableAttribute.tableName,
+                    where = "observation_variable_attribute_name LIKE ?",
+                    whereArgs = arrayOf(name)).toFirst()[ObservationVariableAttribute.PK] as Int
+
+        } ?: -1
+
+        fun getAll() = withDatabase { it.query(ObservationVariableAttribute.tableName).toTable() }
```
BrapiCoordinatorSelby, line 26: unused method

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableDao.kt
```diff
@@ -0,0 +1,349 @@
...
+                    "observation_variable_field_book_format" to "format"
+            )
+
+            val renamedTraits = newTraits.map { renaming[it] }.mapNotNull { it }.toTypedArray()
+
+            renamedTraits
+
+        } ?: arrayOf()
+
+        fun getAllTraitsForExport(): Cursor? = withDatabase { db ->
```
BrapiCoordinatorSelby, line 133: This method never uses the `db` directly, so it doesn't need the `withDatabase` wrapper I think? Unless there is some sneaky black magic I'm not seeing.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableDao.kt
```diff
@@ -0,0 +1,349 @@
...
+                        "validValuesMax" to t.maximum as String,
+                        "category" to t.categories as String,
+                ).asSequence().forEach { attrValue ->
+
+                    //TODO: commenting this out would create a sparse table from the unused attribute values
+//                    if (attrValue.value.isNotEmpty()) {
+
+                        val attrId = ObservationVariableAttributeDao.getAttributeIdByName(attrValue.key)
+
+                        val rowid = db.insert(ObservationVariableValue.tableName, null, contentValuesOf(
```
BrapiCoordinatorSelby, line 262: move this query to `ObservationVariableValueDao`

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableDao.kt
```diff
@@ -0,0 +1,349 @@
...
+                        db.update(ObservationVariableValue.tableName, ContentValues().apply {
+                            put("observation_variable_attribute_value", maximum)
+
+                        }, "${ObservationVariable.FK} = ? AND ${ObservationVariableAttribute.FK} = ?", arrayOf(id, attrId.toString()))
+                    }
+                    "category" -> {
+                        db.update(ObservationVariableValue.tableName, ContentValues().apply {
+                            put("observation_variable_attribute_value", categories)
+                        }, "${ObservationVariable.FK} = ? AND ${ObservationVariableAttribute.FK} = ?", arrayOf(id, attrId.toString()))
+                    }
```
BrapiCoordinatorSelby, line 321: move this query to `ObservationVariableValueDao` and reference it 3 times

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableDao.kt
```diff
@@ -0,0 +1,349 @@
...
+        fun writeNewPosition(column: String, id: String, position: String) = withDatabase { db ->
+
+            db.update(ObservationVariable.tableName,
+                    ContentValues().apply {
+                        put("position", position)
+                    }, "$column = ?", arrayOf(id))
+
+        }
+
+        fun getTraitColumnsAsString() = getAllTraits().joinToString(",")
```
BrapiCoordinatorSelby, line 346: delete unused method

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationVariableValueDao.kt
```diff
@@ -0,0 +1,21 @@
...
+
+        fun getVariableValues(id: Int) = withDatabase { db ->
+
+            return@withDatabase db.query(ObservationVariableValue.tableName,
+                    where = "${ObservationVariable.FK} = ?",
+                    whereArgs = arrayOf("$id")).toTable()
+
+        }
+
+        fun getAll() = withDatabase { it.query(ObservationVariableValue.tableName).toTable() }
```
BrapiCoordinatorSelby, line 19: delete unused method

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+
+            val headers: Array<String> = db.query(ObservationUnitAttribute.tableName,
+                    arrayOf("observation_unit_attribute_name"),
+                    where = "${Study.FK} = ?",
+                    whereArgs = arrayOf("$exp_id"))
+                    .toTable()
+                    .map { it["observation_unit_attribute_name"] }
+                    .mapNotNull { it.toString() }
+                    .filter { it.isNotBlank() && it.isNotEmpty() }
+                    .toTypedArray()
```
BrapiCoordinatorSelby, line 38: Can you use `ObservationUnitAttributeDao.getAllNames()` here instead?

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                GROUP BY units.${ObservationUnit.PK}
+            """.trimMargin()
+
+            println("$exp_id $query")
+
+            println("New switch field time: ${
+                measureTimeMillis {
+                    db.execSQL(query)
+                }.toLong()
+            }")
```
BrapiCoordinatorSelby, line 72: Use a formal `Log` statement, or delete if this was just for debug

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                    .toFirst()
+
+            FieldPreferenceNames(
+                    unique = fieldNames["study_unique_id_name"].toString(),
+                    primary = fieldNames["study_primary_id_name"].toString(),
+                    secondary = fieldNames["study_secondary_id_name"].toString())
+
+        }
+
+        fun getAllStudyModels(): Array<StudyModel> = withDatabase { db ->
```
BrapiCoordinatorSelby, line 99: delete unused method

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                        it.primary_id = cursor.getString(cursor.getColumnIndexOrThrow("study_primary_id_name"))
+                        it.secondary_id = cursor.getString(cursor.getColumnIndexOrThrow("study_secondary_id_name"))
+                        it.date_import = cursor.getStringOrNull(cursor.getColumnIndexOrThrow("date_import")) ?: ""
+                        it.date_edit = cursor.getStringOrNull(cursor.getColumnIndexOrThrow("date_edit")) ?: ""
+                        it.date_export = cursor.getStringOrNull(cursor.getColumnIndexOrThrow("date_export")) ?: ""
+                        it.exp_source = cursor.getString(cursor.getColumnIndexOrThrow("study_source"))
+                        it.count = cursor.count.toString()
+                    }
+
+                } else null
```
BrapiCoordinatorSelby, line 172: Consider using `toTable()` here to mirror the above `getAllFieldObjects()`. 
Then you can make a single convert model to FieldObject method and use it in both cases.

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                db.endTransaction()
+            }
+        }
+
+        private fun updateImportDate(db: SQLiteDatabase, exp_id: Int) {
+
+            db.update(Study.tableName, ContentValues().apply {
+                put("count", db.query(ObservationUnit.tableName,
+                        where = "${Study.FK} = ?",
+                        whereArgs = arrayOf("$exp_id")).count)
```
BrapiCoordinatorSelby, line 281: Consider using `ObservationUnitDao.getAll()` here

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                put("date_import", getTime())
+            }, "${Study.PK} = ?", arrayOf("$exp_id"))
+        }
+
+        private fun modifyDate(db: SQLiteDatabase, exp_id: Int) {
```
BrapiCoordinatorSelby, line 286: `exp_id` is never used... is this method really meant to update ALL studies?

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
BrapiCoordinatorSelby: or even `StudyDao.getCount()`

app/src/main/java/com/fieldbook/tracker/database/dao/VisibleObservationVariableDao.kt
```diff
@@ -0,0 +1,124 @@
...
+
+            db.query(sVisibleObservationVariableViewName,
+                    select = arrayOf("observation_variable_name"),
+                    orderBy = "position").toTable().map { it ->
+                it["observation_variable_name"].toString()
+            }.toTypedArray()
+
+        } ?: emptyArray<String>()
+
+        fun getVisibleTraitObjects(): Array<ObservationVariableModel> = withDatabase { db ->
```
BrapiCoordinatorSelby, line 28: unused method

app/src/main/java/com/fieldbook/tracker/database/dao/VisibleObservationVariableDao.kt
```diff
@@ -0,0 +1,124 @@
...
+                    arrayOf(ObservationVariable.PK,
+                            "observation_variable_field_book_format",
+                            "position")).toTable().map { row ->
+                row["observation_variable_field_book_format"] as String
+            }.toTypedArray()
+
+        } ?: emptyArray()
+
+        //todo switched name to 'getDetails'
+        fun getDetails(trait: String): Array<ObservationVariableModel> = withDatabase { db ->
```
BrapiCoordinatorSelby, line 57: unused method

app/src/main/java/com/fieldbook/tracker/database/models/ObservationUnitPropertyModel.kt
```diff
@@ -0,0 +1,5 @@
+package com.fieldbook.tracker.database.models
+
+data class ObservationUnitPropertyModel(val map: Map<String, Any>) {
```
BrapiCoordinatorSelby, line 3: Is this class meant to be empty? I guess I don't understand the difference between `ObservationUnitProperty` and `ObservationUnitValue` ....

app/src/main/java/com/fieldbook/tracker/traits/PercentTraitLayout.java
```diff
@@ -94,7 +94,11 @@ public void loadLayout() {
             }
 
             setCurrentValueText(loadValue, Color.BLACK);
-            seekBar.setMax(Integer.parseInt(getCurrentTrait().getMaximum()));
+            String max = getCurrentTrait().getMaximum();
+            //TODO: had to add this check, system was parsing empty string as max value which caused an error
+            if (!max.isEmpty()) {
+                seekBar.setMax(Integer.parseInt(getCurrentTrait().getMaximum()));
```
BrapiCoordinatorSelby, line 100: `seekBar.setMax(Integer.parseInt(max));`

app/src/test/java/DatabaseBatchTest.kt
```diff
@@ -0,0 +1,2247 @@
...
+import android.database.sqlite.SQLiteException
+import android.graphics.Bitmap
+import android.os.Build
+import androidx.core.database.getStringOrNull
+import androidx.test.core.app.ApplicationProvider
+import com.fieldbook.tracker.brapi.Image
+import com.fieldbook.tracker.brapi.Observation
+import com.fieldbook.tracker.database.*
+import com.fieldbook.tracker.database.Migrator.Companion.sImageObservationView
+import com.fieldbook.tracker.database.Migrator.Companion.sImageObservationViewName
```
BrapiCoordinatorSelby, line 14: These two constants don't exist anymore... I assume they should be `sLocalImageObservationsView`

app/src/test/java/DatabaseBatchTest.kt
```diff
@@ -0,0 +1,2247 @@
...
+    fun checkGetUserDetail() {
+
+        getAllPlotIds().forEach {
+
+            val oldDetail by lazy {
+                mDataHelper.getUserDetail(it)
+            }
+
+            val newDetail by lazy {
+                ObservationDao.getUserDetail(it)
```
BrapiCoordinatorSelby, line 1328: missing parameter `plotId`

app/src/test/java/DatabaseTest.kt
```diff
@@ -0,0 +1,101 @@
+import android.content.Context
+import android.content.Context.MODE_PRIVATE
+import android.os.Build
+import androidx.test.core.app.ApplicationProvider
+import com.fieldbook.tracker.database.*
+import com.fieldbook.tracker.database.Migrator.Companion.sImageObservationView
+import com.fieldbook.tracker.database.Migrator.Companion.sImageObservationViewName
```
BrapiCoordinatorSelby, line 7: constants missing, should be `sLocalImageObservationsView`

BrapiCoordinatorSelby: reading further, they are unused imports and can be deleted

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
BrapiCoordinatorSelby: TL;DR probably `study_id` is fine, just to keep it separate from the BrAPI specific stuff. 

My personal favorite naming scheme is to match PKs and FKs. So the Study table has PK `study.study_id` and the Obs Unit table would have the FK `obsUnit.study_id`.
My second favorite naming scheme is to FK = table + PK. So the study table has PK `study.id` and the Obs Unit table would have the FK `obsUnit.study_id`.
In this case, that would mean either `internal_id_study` or `study_internal_id_study` which are both fine but pretty wordy. 

Really my issue was only with the confusion with BrAPI fields, since there already is a `study.study_db_id` field but its not the PK, so at first glance it looks like you are matching PKs and FKs, but not really. so you could get away with `study_id` just to keep it separate from the BrAPI specific stuff and have minimal changes to the rest of your code.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
BrapiCoordinatorSelby: My only thought is that lines 148:186 only deal with ObservationVariableAttribute and ObservationVariableValue, and Observations does not depend directly on either of those.

But this just my neurotic code tidiness talking, if it ain't broke don't fix it.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
BrapiCoordinatorSelby: I like DB constraints when possible cuz it stops me as a dev from doing dumb stuff (and it makes ORM libraries super helpful) but I will happily accept that it is an existing Field Book feature and don't break it. Your call.

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+                        put("study_primary_id_name", e.primary_id)
+                        put("study_secondary_id_name", e.secondary_id)
+                        put("experimental_design", e.exp_layout)
+                        put("common_crop_name", e.exp_species)
+                        put("study_sort_name", e.exp_sort)
+                        put("date_import", e.date_import)
+                        put("date_export", e.date_export)
+                        put("date_edit", e.date_edit)
+                        put("study_source", e.exp_source)
+                    }).toInt()
```
BrapiCoordinatorSelby, line 203: Missing `count` from field object

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
```diff
@@ -0,0 +1,346 @@
...
+
+                        put("study_name", e.exp_name)
+                        put("study_alias", e.exp_alias)
+                        put("study_unique_id_name", e.unique_id)
+                        put("study_primary_id_name", e.primary_id)
+                        put("study_secondary_id_name", e.secondary_id)
+                        put("experimental_design", e.exp_layout)
+                        put("common_crop_name", e.exp_species)
+                        put("study_sort_name", e.exp_sort)
+                        put("date_import", e.date_import)
```
BrapiCoordinatorSelby, line 199: The original code set `date_import` to the current time (see DataHelper:2019 commented code)
Either this feature should be maintained, or each calling class needs to be modified to add the import timestamp

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
BrapiCoordinatorSelby: @chaneylc bump

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitDao.kt
BrapiCoordinatorSelby: @chaneylc bump

app/src/main/java/com/fieldbook/tracker/database/dao/StudyDao.kt
BrapiCoordinatorSelby: @chaneylc is this method really meant to update ALL studies?
I was expecting you to use the `exp_id` in the method, not remove it

chaneylc: I changed this back to only edit the selected study.

PR #199, opened 2021-01-20
fix for issue 190 and 196, updated layout params

chaneylc:
# Description

Simple bug fix for issues 190 and 196.


BrapiCoordinatorSelby (approved):
I don't exactly know how to reproduce those issues, so I didn't test it. But the code makes sense and looks good to me.

trife (approved):
From what I can tell this fixed the bug for me

PR #207, opened 2021-02-01
BrAPI updates / bug fixes

nickpalladino:
# Description

- Increased brapi client timeout to 2 seconds
  - Was running into some timeout issues pulling data from breedbase
- Increased BrAPI page size from 1000 to 5000
  - Had a study with 1800 plots which was getting cut off at 1000 so bumped it up, I think this may be being addressed in a better way in new code updates?
- Fixed timestamp format in data helper so that it matches changes made elsewhere 
  - Was getting an exception parsing the last_synced_time timestamps from the db that was messing up brapi export
- Added parsing of BrAPI trait categories in format of value=meaning
  - Only use value part for data stored
  - Add category value=meaning to trait description

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Export via BrAPI and confirm new observations are exported and changed to synced in the statistics
- [X] Import trait via BrAPI with ordinal or nominal datatype and categories in format value=meaning (ex: 0=test cat 1, 1= test cat 2). Verify that available categories on collect screen are shown as 0 and 1. Verify that description has Categories: 0=test cat 1; 1= test cat 2 appended.
- [X] Import study with more than 1000 plots and less than 5000 and verify that all are present on the collect page
- [X] Import study where `GET brapi/v1/studies/{id}/observationunits` takes more than 1 second and less than 2 seconds to complete. Verify that it does not time out.

**Test Configuration**:
* Hardware: Nexus 6P API 23 Emulator
* SDK: Android Studio 4.0.1

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas


trife:
Going to have @chaneylc look over the database modifications and try to get @BrapiCoordinatorSelby to check out the categorical changes.

trife (approved):

app/src/main/java/com/fieldbook/tracker/brapi/BrAPIService.java
```diff
@@ -979,10 +983,30 @@ public void onFailure(ApiException e, int statusCode, Map<String, List<String>>
...
+    }
+
     private String buildCategoryList(List<String> categories) {
         StringBuilder sb = new StringBuilder();
         for (int j = 0; j < categories.size(); ++j) {
-            sb.append(categories.get(j));
+            String value;
+            // parse out only value of specified as according to BrAPI, value=meaning
+            String parts[] = categories.get(j).split("=");
+            if (parts.length > 1) {
```
BrapiCoordinatorSelby, line 1003: You can replace this `if` with `sb.append(parts[0].trim());` 
If `split()` doesn't find a character, it dumps everything into position zero
This is not worth stopping the merge, just a code cleanliness trick.

BrapiCoordinatorSelby (approved):

chaneylc (approved):
Nice, looks good.

PR #208, opened 2021-02-08
Brapi v2

BrapiCoordinatorSelby:
# Description

Implement BrAPI V2 code so Field Book is now compatible with BrAPI V1 and V2. 

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Please describe any tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Test BrAPI field download V1 & V2
- [ ] Test BrAPI observation upload
- [ ] Test BrAPI image upload
- [ ] Test compatibility with Breedbase 
- [ ] Test compatibility with older SDK API versions

**Test Configuration**:
* Hardware: Samsung Galaxy S7
* SDK: 8.0.0 Oreo

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to documentation
- [X] My changes generate no new warnings


BrapiCoordinatorSelby:
Couple of notes on this ^

1. There is a bug in Breedbase code that means I couldn't test it against there. Mirrella is aware of the problem and it is on her todo list
2. because of that, I was not able to test any BrAPI upload functionality. I'm also working on the BrAPI test server to allow auth and upload testing there, but it will take some time
3. I had some concern about SDK API versions, the BrAPI V2 library uses some more advanced java features/libraries. I attempted to test on a virtual device with API 20, but it kept crashing my laptop
4. the new packages are on a GitHub repo. As @Trevor and I discussed this morning, this is annoying cuz you have to provide GitHub creds to download them. I set Gradle up to look for these creds in local.properties and I can provide examples if needed

with all the testing and package shenanigans, this code might not be ready for productions yet ... but its good enough I wanted to get it out there so you guys can see what I was working on and we can resolve conflicts with some of the other recent pull requests

trevor:
hello

chaneylc:
Hey Peter, thanks for this awesome work!

What do you think about swapping some of the FieldBook specific brapi/models for the models in brapi-java-model package?

BrapiCoordinatorSelby:
@chaneylc 
I was trying to keep everything from the BrAPI libraries (v1 and v2) hidden behind the BrAPIService interface. So BrAPIService is responsible for doing the mapping from a BrAPI response object to whatever object Field Book uses. I think ideally the Field Book models should match the Field Book database and can be useful in the rest of the app. With your DB changes, it might be possible to use the brapi-java-models to connect with the DB and be used in the rest of the app, but I'm not sure if that would be particularly helpful. The code to map from the BrAPI model to the Field Book DB has to exist somewhere, my preference is to leave it in the BrAPIService instead of moving it to the DAO classes

chaneylc:
@BrapiCoordinatorSelby 
I implemented a new BrAPI import page in this PR https://github.com/PhenoApps/Field-Book/pull/216 using this v2 branch and have a question about querying with this API.

The focus of the new import is to allow multiple studies to be imported at once, along with searching for trials that belong to many programs (studies that belong to many trials, etc.) Currently, how it works is an initial getPrograms call is used to populate a list, the user can click the programs in the list to select which they will searching within. Once a list of programs are selected a brute-force search for all trials happens like this: 


`mProgramIds.forEach { program ->
  
        mService.getTrials(program.programDbId, mPaginationManager, { trials ->
  
            trials.forEach { trial ->
  
                callGetStudies(brapiStudies, program, trial)
            }
  
            null
  
        }) { fail -> handleFailure(fail) }
    }`


So essentially, to populate a list of all trials within a group of programs you need a p*t (#programs * #trials) number of queries (disregarding paging). This approach works, but I want to make sure it's not overkill, and I would preferably like to minimize network usage in the app.

My question is whether I'm using this wrong or is there a way to do more complex queries s.a



`  val programs = arrayOf(program.programDbId, program2.programDbId)
    mService.getTrials(programs, mPaginationManager, { trials ->

        trials.forEach { trial ->

            callGetStudies(brapiStudies, program, trial)
        }

        null

    }) { fail -> handleFailure(fail) }
  }`

where the query parameters are a list of possible ids.

BrapiCoordinatorSelby:
@chaneylc 
The best way to do this would be to use the `/search/trials` and `/search/studies` endpoints. These endpoint allow you to POST larger search objects, including arbitrarily long lists of DbIds. The code in `BrAPIServiceV2` would look something like this:
```
  BrAPITrialSearchRequest searchRequest = new BrAPITrialSearchRequest();
  searchRequest.setProgramDbIds(programDbIds);
  trialsApi.searchTrialsPostAsync(searchRequest, callback);
```

The down side to this approach is that in the `brapi-java-client` library, these methods are totally untested and I'm willing to bet they are incomplete in some ways. So it can be done, but it might take some extra work in the new library to make it work properly.

BrapiCoordinatorSelby:
I should mention the other downside is the response time. `GET /trials` has an expectation it will be returned as a single call in a "reasonable" amount of time. `POST /search/trials` can potentially be much more complex, which leads to much more complex SQL, and has no guarantees on return time. In fact, there is a polling mechanism built in to detect when results are ready without timing out. 
I don't think this really applies to your case, and my gut says one POST search call will generally be faster than p*t GET calls. I just wanted to call it out so there are no surprises.

chaneylc:
Thanks @BrapiCoordinatorSelby I'll try to implement that call and test the performance against the p*t approach.

BrapiCoordinatorSelby:
I think this is ready to go
it contains the branches `brapi_oauth` and `brapi_v2_hackathon` as well as a bunch of fixes and edits to brapi export 
I also merged in the latest from `develop` so there shouldn't be any major merge conflicts

PR #216, opened 2021-02-19
Issue 172 brapi import

chaneylc:
# Description

This pull request includes a new BrAPI import fragment which is used when users choose "V2" in their preferences. The user can import multiple studies and query across multiple programs/trials. I made small changes to BrapiLoadDialog and BrapiActivity to make this work. I also noticed a potential bug in BrapiPaginationManager where the totalPages was being set to 0 instead of 1 (from my review of the code pages should always begin from 1).

Also included in this PR is a generic view for controlling pages which is used in the new BrapiImportActivity. 

Note about the functionality of the import page:
The user can click the bottom bar to navigate to the next filterable table. If no items were selected this works like a select all for the current page.

Fixes # 172

## Type of change

# How Has This Been Tested?

I've tested this on a Pixel 4a API 30 emulator and moto x4. Because BrapiLoadDialog and BrapiActivity were both changed, I ensured V1 imports still work. 

# Checklist:

https://user-images.githubusercontent.com/5421632/108442261-5bd0e680-721c-11eb-9a18-af46644e967e.mp4


trife:
@chaneylc This looks pretty awesome. A few changes below I'm happy with it:

- Add the toolbar to this activity with a refresh button and a select all (on the current page) button. You can reuse the select all icon from the traits. Clicking the select all button can toggle between select all/deselect all
- Don't close the activity when the user clicks on "programs" since they will have the back arrow on the toolbar.
- How complicated would it be to retain selected state when navigating through the different levels?
- It looks like breedbase hasn't implemented the search services calls (/search/{programs|trials|studies}) which will be an issue. @MFlores2021 can comment on a timeline for supporting these calls but we might need to reapproach this using other calls if they aren't going to be implemented soon.

MFlores2021:
@chaneylc it looks great!. @trife all those search calls are implemented. Let me know if you find any issue. We can take a look.

trife:
There's something weird going on. In both the @BrapiCoordinatorSelby V2 branch where the /search/studies isn't used and this one, cassavabase doesn't load any studies in Field Book. I'm not sure what's going on here, but maybe @chaneylc or Peter can debug. @MFlores2021 both of these branches work on the testserver, so I don't know if this is a breedbase issue or a Field Book issue.

I thought it was related to [this](https://cassavabase.org/brapi/v2/search/studies), that returns `Method GET not implemented for http://cassavabase.org:8080/ajax/brapi/studies_search_save`, but I don't think the other V2 branch is using this call to display studies.

MFlores2021:
@trife: GET search/studies/ requires studyDbId. I'm not familiar with Fieldbook backend, is this using breedbase brapiclient?  brapiclient still working with v1

trife:
@chaneylc we can delay this until v5.1

trife:
@chaneylc I'm closing this for now. We can revisit later this year.

Issue #87, opened 2020-02-21
Design basic fields in Field Book


trife:
When adding a new field, add an option to create a basic field with certain # rows, # columns. Choose starting corner. Add serpentine options for plot number.

PR #218, opened 2021-02-25
Issue 87 field creator

chaneylc:
# Description

Closes issue #87 

PR includes an additional options menu in the field editor activity for generating basic fields. Users input the name of the field, the number of rows and columns, and a plot pattern (linear or zigzag). Plots are enumerated using the pattern and inserted after a review is given, the insert job can be cancelled. 

https://user-images.githubusercontent.com/5421632/109216362-0c3d6e00-777a-11eb-9598-b319b73d10cb.mp4

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Tested on emulator and device. Tested all 8 configurations.

# Checklist:

- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to documentation
- [ x] My changes generate no new warnings


trife:
- Index for both row and column should start at 1 instead of 0
- To clarify, the selection of the corner/direction should be where plot 1,1 exists. The actual plot order is pretty much going to be the same regardless of these selections- the visual layout (when implemented) will be different.
- Is it possible to set the dialog title instead of using different textviews in the layout?
- It's not currently possible to close the creator dialog when opened (back button doesn't work)
- Use "plot_id" instead of "observationUnitDbId" for the user-facing label ; "observationUnitDbId" can still be used internally
- When checking for field already exist, clear out field name (or select name and focus on edittext) but leave the row/column numbers
- Would it be possible to use a viewpager that adds next/back in the dialog creation steps?
- Is it possible to extend from AppAlertDialog instead of Dialog to keep the same theme?

trife:
- Calculating the number of plots is still incorrect. If you design a 10x10 field each row/column only has 81 plots.
- Serpentine plots are being inserted incorrectly. For a 5x5 field, the expected row/col order for a serpentine field would be: {1,1},{1,2}...{1,5}, {2,5}, {2,4}...{2,1}, {3,1}, {3,2}... Since we don't have sort functions for fields, they need to be inserted into the db in the correct order.

trife (approved):

PR #231, opened 2021-03-19
Issue 229

### LINKED ISSUE #229: Import sanitization to prevent SQL injection
When imported headers contain ` or " the app crashes. Instead of trying to catch all of these ( `DataHelper.hasSpecialChars` ), they should just be imported correctly as plain text.

Related to #109 
Related to https://github.com/solgenomics/sgn/issues/3375

chaneylc:
# Description

Uses regex to check for special characters and deletes them. From https://github.com/solgenomics/sgn/issues/3375 will this cause issues downstream?

If we want to accept these with special characters we will have to replace the range table.

Todo: add a message saying that special characters were detected, or replace the range table.

Fixes #229

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Tested sample files with special characters.

# Checklist:

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -29,6 +29,10 @@ android {
                 }
             }
         }
+
+        debug {
+            debuggable true
+        }
     }
 
     compileOptions {
--- a/app/src/main/java/com/fieldbook/tracker/activities/FieldEditorActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/FieldEditorActivity.java
@@ -485,32 +485,32 @@ public class FieldEditorActivity extends AppCompatActivity {
         List<String> list = Arrays.asList(reservedNames);
 
         //TODO causing crash
-        boolean hasSpecialCharacters = false;
-        for (int i = 0; i < importColumns.length; i++) {
-
-            String s = importColumns[i];
-
-            if (DataHelper.hasSpecialChars(s)) {
-
-                hasSpecialCharacters = true;
-
-                importColumns[i] = DataHelper.replaceSpecialChars(s);
-            }
-
-            if (list.contains(s.toLowerCase())) {
-
-                Utils.makeToast(getApplicationContext(),getString(R.string.import_error_column_name) + " \"" + s + "\"");
-
-                return;
-            }
-        }
-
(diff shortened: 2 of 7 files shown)


trife:
Ideally we wouldn't delete the special characters but would just take the proper steps to account for them in the import files- e.g. using argument-based queries instead of raw queries.

chaneylc:
The issue with special characters isn't being caused by raw queries, it's being caused by dynamically generating SQL tables (execSql). The range table is built directly off the imported file's headers which causes the issue. Generally, it's not a good idea to dynamically create tables. A possible alternative is using prepared statements.

trife:
I did some testing with different files and different special characters and it looks like the only ones that actually "break" the creation of the range table building process are ' and ` and ". Others create the table correctly but the app breaks when going to the collect screen with the below logcat. Would it be possible to 1) reduce the replaced special characters to only ' ` " with a notification to the user when these were being replaced and 2) implement the character escapes necessary to support the remainder of the other special characters?


```
        at com.fieldbook.tracker.database.SqlUtilKt.query(SqlUtil.kt:43)
        at com.fieldbook.tracker.database.SqlUtilKt.query$default(SqlUtil.kt:41)
        at com.fieldbook.tracker.database.dao.ObservationUnitPropertyDao$Companion$getRangeFromId$1.invoke(ObservationUnitPropertyDao.kt:56)
        at com.fieldbook.tracker.database.dao.ObservationUnitPropertyDao$Companion$getRangeFromId$1.invoke(ObservationUnitPropertyDao.kt:15)
        at com.fieldbook.tracker.database.SqlUtilKt.withDatabase(SqlUtil.kt:152)
        at com.fieldbook.tracker.database.dao.ObservationUnitPropertyDao$Companion.getRangeFromId(ObservationUnitPropertyDao.kt:50)
        at com.fieldbook.tracker.database.DataHelper.getRange(DataHelper.java:1366)
        at com.fieldbook.tracker.activities.CollectActivity$RangeBox.reload(CollectActivity.java:1720)
        at com.fieldbook.tracker.activities.CollectActivity.onResume(CollectActivity.java:532)
```

PR #259, opened 2021-06-04
Keep signingConfig commented out by default?

nahuel-soldevilla:
# Description

Is is possible to keep the signingConfig commented out by default? 

# How Has This Been Tested?

Debugged using android studio.

# Checklist:

- [ ] I have performed a self-review of my own code

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -32,7 +32,7 @@ android {
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.txt'
-            signingConfig signingConfigs.playStoreConfig //Add your own signing config
+            // signingConfig signingConfigs.playStoreConfig //Add your own signing config
         }
 
         debug {


trife:
Is there a need to do this instead of just compiling with the `debug` build variant?

nahuel-soldevilla:
Thanks @trife, yeah that was basically what I needed, It seems somehow my config was not using any of those variants and there was no warning when I launched the app, but now that I changed it to debug and change it back to release I do see a clear warning:

android-studio-debug-warning

So I changed it to debug and now it's working fine. It seems the `keystore.config` file should be there even for debug, so I created an empty file.
```
Executing tasks: [:app:assembleDebug] in project KSU-FieldBook\Field-Book


FAILURE: Build completed with 2 failures.

1: Task failed with an exception.
-----------
* Where:
Build file 'KSU-FieldBook\Field-Book\app\build.gradle' line: 10

* What went wrong:
A problem occurred evaluating project ':app'.
> KSU-FieldBook\Field-Book\app\keystore.config (The system cannot find the file specified)
```

trife:
I made some changes that should remove the necessity of creating the empty file that I'll push in the next couple days.

Issue #250, opened 2021-05-10
[BUG] BrAPI V2 Service Trait Import not working


chaneylc:
**Describe the bug**
Importing traits through the TraitActivity encounters a NPE in the mapTraits function.
```
E/AndroidRuntime: FATAL EXCEPTION: OkHttp Dispatcher
    Process: com.fieldbook.tracker, PID: 19110
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.brapi.v2.model.pheno.BrAPITrait.getTraitDescription()' on a null object reference
        at com.fieldbook.tracker.brapi.service.BrAPIServiceV2.mapTraits(BrAPIServiceV2.java:756)
        at com.fieldbook.tracker.brapi.service.BrAPIServiceV2.access$700(BrAPIServiceV2.java:64)
        at com.fieldbook.tracker.brapi.service.BrAPIServiceV2$9.onSuccess(BrAPIServiceV2.java:558)
        at com.fieldbook.tracker.brapi.service.BrAPIServiceV2$9.onSuccess(BrAPIServiceV2.java:549)
        at org.brapi.client.v2.BrAPIClient$2.onResponse(BrAPIClient.java:876)
        at okhttp3.RealCall$AsyncCall.run(RealCall.kt:138)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:923)
```
**To Reproduce**
1. Enable V2 and authorize on the test server.
2. Go to Traits from config activity.
3. Go to import -> brapi
4. Crash

PR #278, opened 2021-07-15
issue 250 trait import brapi v2 error

ctucker3:
# Description

The issue was that the brapi server had variables with only a variable name and a `null` in the `trait` field. It looks as if the trait information is required for the field book trait, so I skipped any variable that did not have the `trait` field. 

Fixes # https://github.com/PhenoApps/Field-Book/issues/250

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

I tested it by following @chaneylc bug reproduction steps:

1. Enable V2 and authorize on the test server.
2. Go to Traits from config activity.
3. Go to import -> brapi
4. Crash

You can confirm that the brapi test server has variables that would cause an error with the fix by using an api client to inspect the result of `GET https://test-server.brapi.org/brapi/v2/variables`.

# Checklist:

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/BrapiTraitActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/BrapiTraitActivity.java
@@ -23,6 +23,7 @@ import com.fieldbook.tracker.utilities.Utils;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.BiFunction;
 
 import com.fieldbook.tracker.objects.TraitObject;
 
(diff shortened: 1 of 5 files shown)


trife (approved):

nickpalladino (commented):
Should there be any indication to the user that traits were skipped with an explanation?

ctucker3:
> Should there be any indication to the user that traits were skipped with an explanation?

That seems like a good idea. Maybe something like "Import Success. 4 traits ignore because of missing "trait" field"? What are your thoughts @trife?

trife:
That's a great idea!

ctucker3:
@nickpalladino @trife I added a notification message to indicate to the user how many traits were ignored. I just returned an integer because I didn't want to sink too much time into it for an edge case error. With a little refactoring in the future we could change that Integer into an error validation class that will give specific errors for why something didn't import. Might give more flexibility in showing detailed errors for ignore objects in the future.

Issue #59, opened 2020-02-21
Change advance to next plot/trait missing data


trife:
Expand options to make this either a) across all traits or b) across single traits. User should be able to switch behavior between a/b in the advanced settings.

PR #287, opened 2021-07-21
Added skip entries preference modes.

chaneylc:
# Description

Adjusted behavior preferences to allow three different modes for skipping entries.
Modes:
1. Disabled, no entry is skipped
2. Skip entries that have already observed the active trait.
3. Skip entries that have already observed all visible traits.

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -52,6 +52,8 @@ import android.widget.Toast;
 import android.view.Menu;
 import android.view.MenuInflater;
 
+import com.fieldbook.tracker.database.dao.ObservationVariableDao;
+import com.fieldbook.tracker.database.dao.VisibleObservationVariableDao;
 import com.fieldbook.tracker.preferences.GeneralKeys;
 import com.fieldbook.tracker.preferences.PreferencesActivity;
 import com.fieldbook.tracker.traits.LayoutCollections;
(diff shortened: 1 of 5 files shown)


trife:
@chaneylc expected behavior for mode 3 is that Field Book moves to the next blank plot/trait combination. It currently identifies the correct plot, but does not move to the trait that is missing data.

trife:
Will close #59

trife:
@chaneylc An issue I just ran into with a previous version of the database is that switching the `HIDE_ENTRIES_WITH_DATA` from a Boolean to String is causing a CastException. Probably need to set the previous sharedPref permanently to False and use a new sharedpref name going forward.

chaneylc:
I updated the preference key here https://github.com/PhenoApps/Field-Book/pull/287/commits/7b3c5365dd0c08de0fa75230cc1637277a968c6b so that conflict shouldn't occur.

zrm22:
I have pulled down this branch and I am not sure things are working correctly.  I loaded in the sample data(field_sample) and 2 of the traits: height and flowering.  Then before I enabled the new option I scored the plots with the following table(X means I put a value in):

| Plot      | height | flowering |
| ----------- | ----------- | ----------- |
| 1      |       |        |
| 2   |        |    X   |
| 3   |       |       |
| 4 | X | |
| 5 | | |
| 6 | X | X |
| 7 | | |
| 8 | | X |
| 9 | X | X |
| 10 | X | |

I then move back to plot 1 height and enable the new option.  With this enabled I cannot move outside of Plot 1.  A similar thing happens if I disable the option, move to plot 3 and then re-enable the option.

chaneylc:
@zrm22 My first implementation purposefully would block moving to the next plot if a trait was unobserved on the current plot. With the detail you gave me I changed the implementation to a non-blocking traversal. So even if a trait is unobserved on a plot, the collect activity will move to the next trait for the current plot or to the next plot. Also, if all traits are observed across all plots it will show a message saying "Observations completed" and you won't be able to traverse plots (because you're essentially skipping everything). Thanks for your information and feel free to comment for further changes.

trife:
@zrm22 We'd love to get this integrated into the released version if you get a chance to review it this week.

Issue #303, opened 2021-08-03
[BUG] Selecting "only visible traits" incorrectly exports all traits for database format


trife:
Table format exports correctly.

PR #307, opened 2021-08-04
Fix issue#303 

chaneylc:
Fix for issue #303 
Also added some documentation to the function that was causing issues.


trife:
This radically broke something else in database export and now each recorded observation is applied to every trait regardless of visibility. See attached.
[export.zip](https://github.com/PhenoApps/Field-Book/files/6937820/export.zip)

chaneylc:
This should be fixed now.

trife:
Looks like the "radically broken" feature made it into the v5 release. 😬
I was also accidentally evaluating branch issue304 so your previous fix may have been sufficient.

PR #333, opened 2021-09-24
PR for issue #332 label print update

chaneylc:
Added Zebra Android API jar
Updated print trait layout to use Zebra jar.
This is the proposed fix for issue #332

# Description
You no longer need to have PrintConnect installed to print labels. Although you can still use it if you prefer. Now you can use your device's OS to pair the printer (as you would any other Bluetooth device). 

When collecting print traits, the first time you click the print button (each time you open the collect activity, not for every observation unit) you will see this image which will let you select your printer. The devices shown are paired Bluetooth devices which you will have to manually discover/pair using your OS or the PrintConnect. 

image

The printer status text is no longer shown, but when you press the print button a message will be shown to the user. Messages include: print success, printer paused, printer has no paper, and printer not connected.

A local broadcast manager is used to communicate between the trait layout and the printing thread. When an intent is received with a size it means the print was successful and a database observation will be made.

Fixes # 332

## Type of change


trife (approved):

ramierop:
Hey folks, I'm having some trouble with our Zebra printers and am thinking of getting another model to see if it can work out for us. Will FieldBook work with other mobile printers?

ramierop:
@chaneylc Ok. So, I have a Zebra printer issue. Zebra Printer Setup Utility is not allowing me to change the material from receipt to black mark. And the Zebra apps haven't been updated  for a little over two years, IIRC. Is this an issue that you have encountered?

trife:
@ramierop Unfortunately this is totally outside of our wheelhouse. If you're able to find a different mobile printer manufacturer, we could put in some additional effort to support it, but it won't be before the end of the year.

PR #359, opened 2021-10-28
Issue 355

chaneylc:
Issue #355 fix.


trife:
@chaneylc This still doesn't let the user use the internal GPS if nothing has been paired to Bluetooth. If you delete all of your Bluetooth devices, clicking the pair button in the GNSS layout only shows the `Please pair a device before collecting data` toast

chaneylc:
I just double checked this and I'm not seeing that message, can you check again?

trife:
Sooooo... entirely possible that I might have been testing for this fix on the wrong branch 😄

PR #360, opened 2021-10-29
Added xlsx field file import via local or cloud.

chaneylc:
Used dependencies from https://github.com/andruhon/android5xlsx to parse xlsx files.

Currently if the xlsx file is empty no error or message will be shown.

Also noticed that the apache libraries only use doubles for their numerical cells, I can add a check to convert cells without decimals to integers if that is needed.


trife:
@chaneylc Yeah that double by default does some weird things to fields like row or plot (from the sample files). Is it possible to import everything as strings instead?

PR #367, opened 2021-11-01
change configuration of OAuth library to accept http urls

### LINKED ISSUE #353: [REQUEST] Allow BrAPI OAuth against a server without TLS (http vs https)
**Is your feature request related to a problem? Please describe.**
When debugging, it is useful to connect Field Book to a local BrAPI server. The OAuth/OIDC library that is used in Field Book is configured by default to reject non secured URLs (http). This is a good security feature that prevents users from making a mistake, but it also makes debugging much more difficult. 

@MFlores2021 and @nahuel-soldevilla have both encountered this problem

**Describe the solution you'd like**
The library should be configured to allow http endpoints.
There should be some discussion and consideration about user messaging. Perhaps, if a user enters a URL with http, a warning message should be displayed telling them that their connection will not be secure. 

NOTE
I have code prepared to change the OAuth library configuration. I do not have code prepared for warning the user. I will make a branch for this issue and put the configuration code there so others can decide what to do with it and when to merge it.

trife:
Fixes #353


trife:
@chaneylc can you add the dialog that tells a user that if they enter a URL with http, a warning message should be displayed telling them that their connection will not be secure and rebase this onto develop?

trife:
When evaluating with the test server changed to http, I get the warning but it's followed by an `Error Starting BrAPI Auth` toast. I would assume the expected behavior is a dialog with cancel/proceed options that completes the authentication on proceed?

chaneylc:
We probably need @BrapiCoordinatorSelby to confirm what to do here.
The exception that is causing that error message to display is:
    AuthorizationException: {"type":0,"code":5,"errorDescription":"JSON deserialization error"}
It looks like the auth endpoint doesn't send a valid json response when http is used.

I noticed that if you change the brapi url to http, it also auto changes the oidc discovery url to http in onPreferenceChange (brapi preference). If you change the oidc url back to https the authorization runs correctly. I'm assuming oidc requires https, so we need to remove the onPreferenceChange string replacement.

BrapiCoordinatorSelby:
@chaneylc  I just pushed some new code to this branch that should solve your exception.

I learned something today! There is a method `HttpURLConnection.setInstanceFollowRedirects()`. By its name, it is natural to assume that it would in fact follow redirect HTTP status 3XX. However, it doesn't completely work as expected, especially if you are trying to redirect from `http` to `https` like in our case. 

@chaneylc the error you were seeing is because when you send a request to `http://test-server...` it automatically redirects you to `https://test-server...`. This is a very normal and common practice, but the above mentioned `setInstanceFollowRedirects()` doesn't handle it correctly. So Field Book was getting back something like `<html>301 redirect - This page has moved</html>` and was trying to read that as JSON, causing the "JSON deserialization error".

The code I added should make up for the shortcomings of `setInstanceFollowRedirects()` and automatically redirect correctly if needed. 

Regarding the oidc discovery url, that should be able to be http or https, so I think the `onPreferenceChange` is fine. That oidc url is one of the ones they need to use http during testing.

trife:
@chaneylc Can you review this before it gets merged?

chaneylc (approved):
Looks good, thanks again @BrapiCoordinatorSelby 
From my testing, your code didn't need the setInstanceFollowRedirects() but I only tested on the test server.

PR #370, opened 2021-11-01
issue #349

### LINKED ISSUE #349: Modify InfoBar spinner behavior
Current width in `selector_dropdown` is fixed and changing to `wrap_content` increases width for all items to the max width of the dropdown resulting in wasted space. Implement a dynamic width dropdown. The first proposed solution from the link below works for brapi-imported fields but results in an index out of bounds error with the sample fields.

https://stackoverflow.com/a/40764217

chaneylc:
Dynamic width spinner to save space in info bar recycler view in Collect Activity.

Fixes #349


trife:
Nice! I had used the same code but ran into the crash that you fixed.

Is it possible to cap the dynamic width to 50% of the width of the parent?

PR #375, opened 2021-11-03
fixes issue #290

### LINKED ISSUE #290: [BUG] Delete observation no longer shows N/A value
**Describe the bug**

I think it used to be the case that when you exported a brapi observation and then deleted it in the collection activity the value was set to "N/A" and a message was displayed telling you why that happened. 

This behavior is missing in v5.0 and v4.5 from what I can tell. 

I haven't looked back in versions to see where this was lost. 

**To Reproduce**
1. Import BrAPI studies and traits
2. Collect some data
3. Export that data via brapi
4. Go back to the collection screen
5. Delete one of your observations
6. The field will go blank and there will be no message

**Expected behavior**
N/A value is put into field and a warning message is displayed informing the user that brapi observations can't be deleted. Using N/A instead. 

**Screenshots**

 - Model: Pixel 3a
 - OS: Android 10
 - Field Book Version v5.0 and v4.5

chaneylc:
bugfix where timestamp wasn't queried when creating brapi observation object
bugfix within isBrapiSynced that didn't return correct boolean
bugfix in validate data Collect Activity

Fixes #290

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -341,6 +341,9 @@ public class CollectActivity extends AppCompatActivity implements SensorEventLis
 
     /**
      * Is used to ensure the UI entered data is within the bounds of the trait's min/max
+     *
+     * Added a check to return NA as valid for BrAPI data.
+     *
      * @return boolean flag false when data is out of bounds, true otherwise
      */
     private boolean validateData() {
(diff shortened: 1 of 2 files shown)


trife:
@ctucker3 Could you evaluate this fix next week to determine if it solves the issues you observed?

ctucker3 (approved):
Sorry about the delay on reviewing this! I ran through a test run on the brapi test server and it looks like this fixes the issue from #290.

PR #377, opened 2021-11-03
Issue109

### LINKED ISSUE #109: Data sanitation issues with bad import files
- Remove extra empty rows/cols

chaneylc:
When importing a field file, the minimum required information is the primary, secondary and unique identifiers. This commit adds a check to the import that skips the line if these are empty.

Fixes #109


trife:
I'm unable to import this file with some extra rows and columns added in. I think it interprets the empty unique IDs as duplicates.

[field_sample_sanitation.csv](https://github.com/PhenoApps/Field-Book/files/7470806/field_sample_sanitation.csv)

chaneylc:
Good catch, should be good now.

PR #379, opened 2021-11-04
Improved quick goto search.

### LINKED ISSUE #157: [BUG] Quick GoTo bug(s)
- Field Book searches in reverse for the first occurrence of the search term
- Current implementation only allows the user to search for either a primary or secondary and not both simultaneously

Need to verify that these are both still problems

chaneylc:
QuickGoto search now uses both primary/secondary terms if they are changed. Otherwise it does the default search with the last changed term.

Fixes #157


trife:
Should we get rid of the listeners on line 2388/2389?

chaneylc:
It looks like we can.

chaneylc:
After testing this I think we should keep the on touch listeners. Without them, the edit text cursor does not appear.

PR #381, opened 2021-11-10
Issue 347 GeoNav Updates

### LINKED ISSUE #347: GNSS/GeoNav Updates
- Modify internal GPS decimal places to use internal precision
- Change snackbar colors to not be black background

chaneylc:
Created high contrast snack bar message for geo navigate
GeoNav log:
Now prints unique id.
Fixed gps updates and iz log race condition
Change column order to show start lat/long first and plot parameters (primary/secondary/unique) last
Added third state to closest (look at documentation for more detail)
Added trapezoid method
Added separate preferences page for geonav
removed triangle method

Fixes #347


trife:
This looks great!

- Activity title in toolbar says `Behavior` instead of `GeoNav`
- `GeoNav` preference section header can be removed
- Change `Search Method` summary to show selected method. If we can include it in the current summary to be a cohesive narrative that would be great (i.e. `The %s method is being used to search for plots`)
- Add units to `Trapezoid` parameter settings.
- Add current values to summaries for all parameters (see above) and update interval.

PR #382, opened 2021-11-10
issue 247 neutral edit text pref

### LINKED ISSUE #247: Create custom EditTextPreference that includes neutral button with barcode scanner text input
- Use for OIDC URL
- Use for BrAPI Base URL

chaneylc:
Updated brapi preference url and oidc url with barcode scan button.

Fixes #247


trife:
@chaneylc The new edittext broke the `Base URL` summary.

trife:
I'm getting some weird behavior when scanning barcodes. I'm not sure if the contents of the barcode scanner are being parsed incorrectly or what. I'm getting a lot of 404's when attempting to use it to scan this barcode... but not always: 
image

Steps to reproduce:
1. Select CassavaBase from community servers (this is also giving a 404 but separate issue)
2. Scan attached barcode from the Base URL preference
3. 404?

chaneylc:
I fixed this based on our discussion @trife 
One thing to note is that when you choose a community server it will automatically change the authentication type. This is not true for when you scan a barcode, since (from my knowledge) there's not an easy way to check if that url is a discovery url. So the workflow would look like:
1. select cassavabase
2. scan barcode
3. change auth type
4. authenticate

For that reason I removed the automatic authenticate after scanning a url. Otherwise it would result in a 404 before the user changes the authentication type.

PR #395, opened 2022-02-04
Allow for saving of data via BrAPI to be chunked

timparsons:
# Description

The original BrAPI implementation attempted to send all observations in a single BrAPI call for both PUT and POST endpoints.  A single request could be too large, and therefore not saved at all.  This change chunks up the save operation into batches of 500 observations, and submits at most 2 at a time until all chunks have been persisted.

The BrAPI export screen has been updated to show the progress of the chunked saving in real time by updating the number of observations that have been sync'd as well as the number of observations that are not sync'd.

A GitHub Action was also created, which will run the unit tests whenever a pull request is opened or edited, as well as whenever code is committed to the `develop` or `master` branches.

## Type of change

# How Has This Been Tested?

I had a db export from a client where they had recorded over 14k observations.  I had a local Breedbase server running with the required study and observationVariable configuration.  To test, I loaded the db file, then went through the BrAPI export screen, and exported the data to my local Breedbase instance.

I also updated the `BrapiServiceTest` unit test to test both BrAPI v1 and v2, and added new test cases for saving observations in chunks.

**Test Configuration**:
* Hardware: Pixel C emulator
* SDK: 10

# Checklist:


chaneylc:
This looks good so far, first of all thank you for the contribution. I tested it out on the brapi test server and everything seemed to work similarly to the develop branch, and I wasn't able to test with a large data set like you have.

Two things I noticed that we can discuss:
`        return createChunks(items, 500); //TODO use pageSize for chunk?
`

You mention using pageSize potentially here, which is already a defined user preference. I think it's good to separate the two and make chunk its own preference. Do you have thoughts on making this user defined, or was there a specific reason for using 500? If you want I can implement the preference, but let me know if there should be any constraints on the integer other than 0 or max int.

`        Semaphore requestSemaphore = new Semaphore(2);
`

Similarly for the parallel writes, you mention changing this from two to being based on device specs. What are you thoughts on just making this another preference?

timparsons:
@chaneylc 

> This looks good so far, first of all thank you for the contribution. I tested it out on the brapi test server and everything seemed to work similarly to the develop branch, and I wasn't able to test with a large data set like you have.
> 
> Two things I noticed that we can discuss: ` return createChunks(items, 500); //TODO use pageSize for chunk?`
> 
> You mention using pageSize potentially here, which is already a defined user preference. I think it's good to separate the two and make chunk its own preference. Do you have thoughts on making this user defined, or was there a specific reason for using 500? If you want I can implement the preference, but let me know if there should be any constraints on the integer other than 0 or max int.

Good catch! I'm happy to have you implement something for this.  I don't see a reason to not have it be a user preference.

> 
> ` Semaphore requestSemaphore = new Semaphore(2);`
> 
> Similarly for the parallel writes, you mention changing this from two to being based on device specs. What are you thoughts on just making this another preference?

It may be easier for now to leave it as two, and give yourself some time to research how best to determine how to dynamically set the number of threads based on device specs.

chaneylc (approved):

PR #417, opened 2022-04-25
Fix for importing files with duplicate columns not populating range table correctly

chaneylc:


trife:
- Add notification to the user that only the first instance of the column name will be imported

PR #418, opened 2022-04-27
[BRAPI] Enabling repeated measures

HMS17:
# Description

Initial steps for implementing subobservations in the form of repeated measures. Having the ability to record multiple observations for a single observation variable/observation unit pair will be greatly helpful for collection in cases like time series data.

**Most Recent Commit**

- "Enable Repeated Observations" is added as a new checkbox option in Settings/Behavior.
- An add repeated measure button has been added to the collect screen, which shows if "Enable Repeated Observations" has been checked.
- When "Enable Repeated Observations" is selected, adding and deleting observations is done on a by-replicate basis, so deletion will only delete one observation for an observation unit/observation value pair. 
- When plus button clicked, new entry in observations table created utilizing existing rep column, collection will now be done on this new repeated observation

**Notes**

- "Enable Repeated Observations" uses the same icon image as "Cycling traits advances entry" but as the file the latter uses "@drawable/ic_adv_cycle_traits_advances" is functionality specific, for "Enable Repeated Observations" a different import of the same icon was made under the more generic "@drawable/ic_repeat".
- Currently the start of a display of the replicate number in activity_collect.xml but ran out of time to retrieve replicate number.
- This PR also fixes a minor bug where there wasn't a default value for label/value selection in settings.

**Future Steps:**

- On collect page for a given observation unit/observation variable pair:
-- Display current replicate number
-- Display to toggle through displaying/editing different repeated observations 
- Improve refreshing of observation display on new replicate
-- Category should deselect
-- Date should reset
-- Finding some way to refresh would be preferable to piecemeal resetting each value type
- Handling case where current behavior for non-text based values is that after a replicate is made, selecting a new value for the replicate somehow triggers a removeTrait and thereby deletes the most recent replicate and instead updates the replicate before that
- Determine how to handle user toggling off repeated observations and effect on database
- Import/Export from/to BrAPI, utilizing "observationlevels"/"observation hierarchy"
- May want to reconsider text functionality of deleting observation on empty string when replicates active?

**Commit 1:** documented in case need to go back to simplest commit
"Enable Repeated Observations" is added as a new checkbox option in Settings/Behavior. When "Enable Repeated Observations" is selected, each collection activity will generate a new replicate and most recent value will be displayed. When "Enable Repeated Observations" is deselected, collection activity will overwrite all replicates and replace them with one entry.

- This leads to two problems:
- deselecting a value counts as a collection activity and generates a replicate with a blank value. selecting a new value generates another replicate, so we have the blank value replicate stuck there
- each keystroke for a text value generates a new replicate, so we get "t" "te" "tes" "test" for trying to enter "test"

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

Please describe any tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] "Enable Repeated Observations" is visible under Settings/Behavior and can be enabled/disabled
- [ ] Plus button on Collect screen is visible when "Enable Repeated Observations" is selected and not visible when deselected
- [ ] When plus button is clicked, a new entry in observations table is added with a rep value one greater than the previous highest rep for that observation unit/observation value combo and a blank value.
- [ ] When plus button is clicked, text should clear
- [ ] When user selects a category after creating a replicate, the value for the new replicate should be the one that updates _**(working earlier, presently failing as it seems it triggers a stray removeTrait and removes the most recent replicate and then updates the replicate before that, see behavior when more than two replicates. ran out of time to test alas especially with slow debugger on collect )**_
- [ ] When user types text after creating a replicate, the value for the new replicate should be the one that updates

**Test Configuration**:
* Hardware: Pixel 2 API 30
* SDK: Android Studio

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings


HMS17:
Moved to cleaner [brapi-subobservations2](https://github.com/PhenoApps/Field-Book/tree/brapi-subobservations2) branch

PR #438, opened 2022-05-16
issue 434

### LINKED ISSUE #434: [BUG] Add Bluetooth permission check to GNSS trait

chaneylc:
# Description
added runtime bluetooth permission checks for android 31+
Fixes #434


trife:
@chaneylc On my test device, this shows in Collect regardless of which traits have been created- I'm using the sample data. If I decline it will show with the GNSS trait connect button as expected. Can we isolate its appearance to the specific traits it's needed for?

Issue #443, opened 2022-05-17
GeoNav: snackbar changes


trife:
- Change duration to permanent using `LENGTH_INDEFINITE`
- If newContents==oldContents, don't reshow
- If newContents!=oldContents, redraw
- Increase height/text size by 10%

PR #456, opened 2022-06-01
updated geonav snack bar navigate layout

chaneylc:
issue #443


trife:
- Custom snackbar isn't taking up the full width of the screen. It looks like there's some additional hanging background somewhere- there are two blobs of colorPrimary on each side.
- I adjusted the layout a little bit. Go ahead and make the icons/text white and bold the text.
- If the user switches to the plot being suggested via nav arrows the snackbar should go away. Maybe redo the search if manual navigation is performed?

Issue #444, opened 2022-05-17
Improve performance when navigating photo trait


trife:

trife:
Potentially related to https://github.com/PhenoApps/Field-Book/issues/90

Issue #90, opened 2020-02-21
Use thumbnails for photo trait data display


trife:
Related to #76

PR #459, opened 2022-06-16
Updated photo trait to display thumbnails, removed async task.

chaneylc:
issues #90 #444 
Doesn't use new camera api yet.


trife:
- I think the thumbnails are getting rotated before being displayed
- Nest thumbnails inside of the photos directory and ensure that they aren't added to the phone's image directory (maybe by adding a .nomedia file? I can't quite remember)
- I'm getting really bad lag when navigating to plots with or without photos captured. Is this doing a full lookup in the DocumentTree that's slowing things down? The below is from logcat while navigating:

`2022-07-11 02:22:36.400 13035-13035/? D/DocumentTreeUtil: Checking volume uri content://com.android.externalstorage.documents/tree/primary%3AfieldBook: null`

trife:
This works great! Only change needed is to nest the thumbnails folder inside of the photos folder and make it hidden by naming it `.thumbnails`. The aspect ratio of the thumbnails doesn't match the original images, but that's not super important.

This PR may interfere with the USB cam trait PR quite a bit, so might make sense to merge them into a single PR?

PR #461, opened 2022-06-22
updated locking feature to locked/unlocked/frozen

chaneylc:
Added three states to lock.
Unlocked/Locked - normal
Frozen - no data can be edited

I can change this so default values can be edited, comment if that's what you prefer.
Also, traits with default values will be automatically saved to the database when navigated to. I can change this to be trait dependent.


trife:
The current approach enables the current value editText, which results in the keyboard being displayed when the user clicks on it.

trife:
@zrm22 If I remember correctly, this was some functionality that had previously been requested by your group and may be of interest.

PR #467, opened 2022-06-28
updated categories backend and trait creator

chaneylc:
Updated trait editor to remove the delimiter requirement.
Old categories are still accepted in case they exist in the database.


trife:
- In the creator, I think it's safe to assume that all labels will be the same as the values, so let's only ask the user to add the different values.

trife:
- Replace Add/Delete with icon buttons
- Move Add button to the right of the new category edittext

PR #468, opened 2022-06-29
TTS

chaneylc:
Added system TTS, speech triggers within the collect activity, generally when an observation is updated/inserted into the database or a button is pressed, excluding the audio trait.


trife:
This works great! Few changes below:

- Add speech actions bottom toolbar actions (Barcode scan → "Opening barcode scanner", NA → "Not applicable", Delete → "Value has been deleted"
- In preferences, close language selector screen once language has been selected. Update the language summary with the current selected value.
- For numeric, categorical, multicat, and date, read out the button presses instead of the current value. Especially important for multicat since each additional category added results in the entire string being read.
- Disease trait reads out button pressed but skips the built in catches (e.g. severity already scored toast). For these catches, the value of the toast can be read aloud.
- Location/GNSS can be switched to "Current location saved" instead of the coordinates
- For percent, don't read value while the slider is still being manipulated
- For calendar, can we read just day/month instead of full day?

trife:
This is amazing!

- Let's move both of these preferences to a new section in the Audio preferences.

trife:
- Replace custom fragment for language selection with listpreference
- Remove toast when selecting language

Issue #422, opened 2022-05-06
[TRAIT] USB Camera connection 


trife:
e.g. https://smile.amazon.com/gp/product/B07PBHM7KV/ref=ppx_yo_dt_b_asin_title_o04_s00?ie=UTF8&th=1

PR #475, opened 2022-07-26
V5.3/features/422 USB Camera trait

chaneylc:
issue #422 

Tested and working on Pixel 3 (9) and Pixel 6 (12)

Does not work on Pixel 4a (causes reboot)


trife:
- Replace `Pair Device` button with circle button using [this icon](https://materialdesignicons.com/icon/connection)
- Let's make the preview look closer to something like this with expand and capture buttons included as part of it.
preview
- Expanding the preview should go to full screen and look something like this: 
Screenshot_20220801-205142
- Captured images should be included in an array similar to the current photo trait. The preview can be one of the items in the array.
- Captured images aren't be collected with the correct aspect ratio. Not sure if there's a reliable way to fix this.
- Move the USB storage folder up one directory so that it's `~/plot_data/<trait_name>/`. Thumbnails should be nested within each of these trait-specific folders.

trife:
This looks awesome!!!
Minor changes:

- Let's move the previous captures below the preview. At some point we might need two layouts- one for horizontal images and one for vertical. I'm fine with a single one for now.
- Scroll the previous captures to the most recent capture (currently stays on the first).
- If the connection to the camera is interrupted while in the Collect screen, the preview never goes away. Should we add a check that the connection is still valid either 1) when the trait is redrawn or 2) when the preview surface is clicked?

chaneylc:
I'm not seeing any missing strings, has that been fixed for you?

trife:
I think it was Android studio weirdness. I'll get this reviewed and merged today.

trife:
@ch728 this will be in the 5.3 beta we (hopefully) release next week.

PR #479, opened 2022-07-29
v5.3/features/language_selection/168

### LINKED ISSUE #168: Reconfigure language selection
Currently requires user to change system language. Allow user to override system language within the app.

https://github.com/akexorcist/Localization

chaneylc:
# Description

Added app-specific language selection in Settings/Appearance/Language
Added non-translatable with combined translatable language list selection.
Uses preferences to persist language selection on app start-up.
Added locale_configs.xml to be used within system settings for android 33.
Updated gradle.
Updated app compat dependency.

Might need to revisit some basic app layout design to better suit RTL languages s.a Arabic.

Fixes #168


trife:
- Is it possible to use a preference page for the language selector screen instead of the custom layout?
- (will be solved if the above is changed) hitting the back button without selecting a language doesn't remove the layout
- Add flags as the icons for each language. You can look at release 2.0 source for some/most of them
- Add default option to use the system language
- Don't need the English translation for each language.
- Alphabetize languages in list
- Switching from Arabic to another language doesn't reverse the RTL layouts until prefs are closed

chaneylc:
Updated. I couldn't reproduce your last bullet though, let me know if that is still occurring for you.

trife:
- Still experiencing the inconsistent RTL behavior, but it's minor and goes away once the pref activity is closed
- Selecting language goes back to the main settings screen instead of Appearance. If there's a technical reason for this, it can stay this way.
- Remove `Language` category title from `preferences_language.xml`

chaneylc:
I'm not sure why its going back two fragments...

trife:
- If you select the current language (e.g. English), the fragment doesn't react.
- I think the summary in the Appearance section is hard-coded.

PR #486, opened 2022-08-04
hotfix/brapi_preference/485

### LINKED ISSUE #485: [BUG] Different page was displayed after authorizing while OAuth2 Implicit Grant was selected
**Describe the bug**
The user is always redirected to BRAPI Test Server ([auth.brapi.org](http://auth.brapi.org/)) when use 'OAuth2 Implicit Grant' as OIDC Flow.

**To Reproduce**
1. Go to Settings -> BrAPI 
2. Tap OIDC Flow and then select OAuth2 Implicit Grant.
3. Tap Base URL and enter a base url
4. Tap Save and Authorize

**Expected behavior**
User should see the login page set as the base URL

 - Model: Pixel 2
 - OS: Android 28
 - Field Book Version: v5.2
 - 
**Additional context**
I guess the problem lies in this line: https://github.com/PhenoApps/Field-Book/blob/develop/app/src/main/java/com/fieldbook/tracker/activities/BrapiAuthActivity.java#L118.

chaneylc:
# Description

OIDC url was not updating correctly when preference was saved.

Fixes #485

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/preferences/BrapiPreferencesFragment.java
+++ b/app/src/main/java/com/fieldbook/tracker/preferences/BrapiPreferencesFragment.java
@@ -258,9 +258,10 @@ public class BrapiPreferencesFragment extends PreferenceFragmentCompat implement
 
     private void updateUrls(String newValue) {
         SharedPreferences sp = prefMgr.getSharedPreferences();
-        String oldBaseUrl = sp.getString(GeneralKeys.BRAPI_BASE_URL, "");
-        String oldOidcUrl = sp.getString(GeneralKeys.BRAPI_OIDC_URL, "");
-        String newOidcUrl = oldOidcUrl.replaceFirst(oldBaseUrl, newValue);
+        String newOidcUrl = newValue + "/.well-known/openid-configuration";
+        if (newValue.endsWith("/")) {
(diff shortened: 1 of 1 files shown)


BrapiCoordinatorSelby:
I don't have my computer setup to test this right now, but just reading the code, I think these changes will introduce a new bug. I think you will have an issue if someone tries to edit the OIDC URL directly. Because of the added `updateURLs()` call on line 389, if someone edits the OIDC URL to `test.com/oidc_conf.json` then that value will be assigned as the base URL, and the OIDC URL will be changed to `test.com/oidc_conf.json/.well-known/openid-configuration`, and both of those values are incorrect.

The way I had it setup originally, if the Base URL and the OIDC URL shared the same host/root-path then an update to the Base URL would modify the OIDC URL to match. This was the default behavior since they started as matching. However, if a user explicitly modified the OIDC URL to have a different host/root-path, then updates to the Base URL would not override that setting. Example of this would be if you were using an external provider for authentication (ie github, google, etc) the OIDC URL would not match the Base URL and you don't want to have to change it every time the Base URL changed. 

My original code was in `onPreferenceChange()` which I think is now unused and could be deleted now that the new preference pop-ups are being used. You don't have to keep my original logic, just be aware that there are very valid use cases where the OIDC URL will be different from the Base URL,

PR #504, opened 2022-09-16
Themes

### LINKED ISSUE #98: Add theme options
- Saved data color scheme: ~~color choice~~, invert background
- App theme: default, eInk/high contrast, blue, monochrome
- Font size

chaneylc:
# Description

Added themes

Fixes #98


trife:
All testing on Pixel 3, Android 12

- Did we get rid of accent color? I liked the contrast of the brown
- The following are two small: create trait list of trait formats (traits), infobar prefix and content, trait name (collect), trait dropdown (collect)
- Top bar spacing is off when an overflow menu is present (trait activity)
- Toast is missing Field Book icon
- Summary text for settings currently matches style of title. Should be smaller and slightly more gray (check specifications).
- Shift text size names: Extra Small → Medium, Medium → Large, Large → Extra Large
- Remove Extra Extra Large and Extra Small
- Add new, smaller values for Small
- Medium should be the default setting and indicated as default in the list. Where possible, Medium should be what the current app looks like prior to themes being added.
- Restore default theme should reset size to medium
- Experiencing a bug with locking where I can't enter new data when set to phase 2 and moving backwards through field (not in 5.3 on Pixel 6)
- High contrast theme should 1) only use white/black
- Changelog still uses old dialog style
- Libraries activity crashes
- I'm not a big fan of the stacked dialog options used in Base URL. Maybe switch the option tags to "Scan", "Cancel", "Authorize" to keep on a single line

trife:
- Textview curCount can be removed (counter)
- Audio text does not match cvET format
- Audio text does not show "NA" until user navigates from/to current plot
- Possible to make the text inside of the multimeasure text box movable to see especially large strings?
- Summary page does not match current theme (hardcoded green)
- High contrast theme needs to change color of bottom toolbar buttons to black
- Observations dialog (multimeasures) does not apply existing themes (as far as I can tell)
- Add animation for multimeasures when deleting observation

trife:
Superseded by #537

PR #511, opened 2022-10-21
V5.4/features/multiple measurements/65

### LINKED ISSUE #65: Multiple measurements for entry/trait
image

chaneylc:
# Description

Added a new way to input data to allow users to enter multiple observations for each plot/variable pair.

Enable by navigating to Settings/Beta Features and ensure the repeated values feature is checked.

Fixes #65

@timparsons @BrapiCoordinatorSelby 
I made some slight changes to the Brapi export activity to allow multiple observations to be pushed. Now, the uniqueness of observations is based on FB's internal db id, and not the plot id and trait id. It would be great if you could test this out a bit. I was able to push multiple observations to the test server, but the response did not return with db ids.


BrapiCoordinatorSelby:
@chaneylc A while ago I think we talked about using `ExternalReferences` as a way to pass DbIds back and forth. (maybe I didn't talk about this with you, could have been someone else, I don't remember)

The server is responsible for setting a DbId when you upload something. Even if you try to send observations with a DbId from FB, the server will ignore it and generate its own. So, if you changed the uniqueness of an observation to be based solely on FB DbId, then uploading it would lose that id as you have described.
Somewhere ([BrAPIServiceV2.java line 668](https://github.com/PhenoApps/Field-Book/blob/4f922c4216098fa7dd6d0b8351815c30dfd95cc5/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java#L668)) you need to add some code that moves the FB DbId into the `externalReferences` array for upload. Then ([BrAPIServiceV2.java line 638](https://github.com/PhenoApps/Field-Book/blob/4f922c4216098fa7dd6d0b8351815c30dfd95cc5/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java#L638)) you need to pull the FB DbId back out of `externalReferences` to be used for uniqueness checks.

Basically, when its all done uploading, you should have one copy of the observation in FB
```
  {
    "observationDbId": "FieldBookID-123",
    "externalReferences": [{
      "referenceSource": "https://test-server.brapi.org",
      "referenceId": "TestServerID-ABC",
    }],
  ...
}
```

And another copy of the observation on the server
```
  {
    "observationDbId": "TestServerID-ABC",
    "externalReferences": [{
      "referenceSource": "Field Book Upload",
      "referenceId": "FieldBookID-123",
    }],
  ...
}
```
That way both FB and the Server have a the data, and they have references to each others copies

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -2021,6 +2090,87 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+
+                int n = values.length;
+
+                if (n == 1) {
+                    if (values[0].getValue().isEmpty()) {
+                        n = 0;
+                    }
+                }
+
+                switch (n) {
```
BrapiCoordinatorSelby, line 2130: Maybe consider some code simplification here. I looked up how to do this, the code is partly from SO and completely untested, but maybe something like this:
```
    String drawableName = "numeric_" + n + "_box";
    int drawableID = this.getResources().getIdentifier(drawableName, "drawable", this.getPackageName());
    item.setIcon(drawableID);
```

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -2185,7 +2331,7 @@ public UsbCameraHelper getCameraHelper() {
 
     ///// class TraitBox /////
     // traitLeft, traitType, and traitRight
-    private class TraitBox {
+    public class TraitBox {
```
BrapiCoordinatorSelby, line 2334: Best Practice: public classes should have their own java file.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -467,27 +466,44 @@ private void initToolbars() {
 
         deleteValue = toolbarBottom.findViewById(R.id.deleteValue);
         deleteValue.setOnClickListener(v -> {
+            boolean status = dt.isBrapiSynced(getStudyId(), getObservationUnit(), getTraitName(), getRep());
             // if a brapi observation that has been synced, don't allow deleting
-            String exp_id = Integer.toString(ep.getInt(GeneralKeys.SELECTED_FIELD_ID, 0));
-            TraitObject currentTrait = traitBox.getCurrentTrait();
-            if (dt.isBrapiSynced(exp_id, rangeBox.getPlotID(), currentTrait.getTrait())) {
-                if (currentTrait.getFormat().equals("photo")) {
+            if (status) {
```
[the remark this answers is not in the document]
BrapiCoordinatorSelby: I think its fine cuz its using `brapiDelete()` instead of `deleteTraitListener()` 
It would be good to double check though

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
```diff
@@ -221,11 +222,11 @@ public void updateTraitVisibility(String trait, boolean val) {
      * this function as well
      * v1.6 - Amended to consider both trait and user data
      */
-    public long insertUserTraits(String rid, String parent, String trait, String userValue, String person, String location, String notes, String exp_id, String observationDbId, OffsetDateTime lastSyncedTime) {
+    public long insertUserTraits(String rid, String parent, String trait, String userValue, String person, String location, String notes, String exp_id, String observationDbId, OffsetDateTime lastSyncedTime, String rep) {
 
         open();
 
-        return ObservationDao.Companion.insertUserTraits(rid, parent, trait, userValue, person, location, notes, exp_id, observationDbId, lastSyncedTime);
+        return ObservationDao.Companion.insertUserTraits(rid, parent, trait, userValue, person, location, notes, exp_id, observationDbId, lastSyncedTime, rep);
```
BrapiCoordinatorSelby, line 229: Consider changing the names of these methods like you have done elsewhere `insertUserObservations()`

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
```diff
@@ -246,14 +268,36 @@ class ObservationDao {
...
                     ObservationUnit.FK to rid,
                     ObservationVariable.FK to internalTraitId
             ))
 
         } ?: -1L
 
+        fun insertRep(rid: String, parent: String, trait: String,
+                             userValue: String, person: String, location: String,
+                             studyId: String, rep: String): Long = withDatabase { db ->
+
```
BrapiCoordinatorSelby, line 282: Why do you need a separate method specifically for reps? Doesn't `insertUserTraits()` do the same thing?

app/src/main/java/com/fieldbook/tracker/traits/BaseTraitLayout.java
```diff
@@ -113,31 +195,60 @@ public SharedPreferences getPrefs() {
...
+        if (inputView.isRepeatEnabled()) {
+            repeatView.setVisibility(visibility);
+        } else {
+            editText.setVisibility(visibility);
+            editText.setHint("");
+
+            if (isTraitType(TextTraitLayout.type)
+                    || isTraitType(AudioTraitLayout.type)
+                    || isTraitType(PhotoTraitLayout.type)) {
+                editText.setVisibility(View.GONE);
```
BrapiCoordinatorSelby, line 237: Why is `TextTraitLayout` included in this OR? Elsewhere in the code `TextTraitLayout` is being explicitly calling `toggleVisibility(View.VISIBLE)`, then this negates that if repeat is disabled?
Also should `UsbCameraTraitLayout` to match the OR statement on line 78?
Also are the two OR statements (lines 78 and 234) redundant?

app/src/main/java/com/fieldbook/tracker/traits/CategoricalTraitLayout.java
```diff
@@ -45,6 +45,13 @@ public CategoricalTraitLayout(Context context, AttributeSet attrs, int defStyleA
         super(context, attrs, defStyleAttr);
     }
 
+    static public boolean isTraitCategorical(String traitName) {
+        for (String name : POSSIBLE_VALUES) {
+            if (name.equals(traitName)) return true;
+        }
+        return false;
+    }
+
```
BrapiCoordinatorSelby, line 54: Nothing technically wrong with this method ... but I hate it
- having a static method in a random layout class seems wrong 
- `traitName` should be `traitFormat`
- `POSSIBLE_VALUES` should be something like `CATEGORICAL_FORMATS`
- The _one_ place where this method is used in `CollectActivity:1000` could have just as easily had the line 
```
Arrays.toList(POSSIBLE_VALUES).contains(traitFormat)
```

If I were to fix this properly (probably in a different pull request) I would create an Enum for all the supported traitFormats. Then in the Enum you could add flags like `isCategorical()`. Then it would just be `traitFormat.isCategorical()`

Sorry for my little tantrum. Like I said, there is nothing technically wrong, it will work. Just something to keep in mind for a later refactoring.

app/src/main/java/com/fieldbook/tracker/traits/LabelPrintTraitLayout.java
```diff
@@ -200,7 +200,7 @@ public void init(Activity act) {
     @Override
     public void loadLayout() {
 
-        getEtCurVal().setVisibility(EditText.GONE);
+        getCollectInputView().setVisibility(EditText.GONE);
```
BrapiCoordinatorSelby, line 203: Should this also use `toggleVisibility()`? Should `EditText.GONE` be `View.GONE`?

app/src/main/java/com/fieldbook/tracker/views/CollectInputView.kt
```diff
@@ -0,0 +1,120 @@
...
+            repeatView.text
+        } else editText.text.toString()
+
+        set(value) {
+            if (repeatModeFlag) {
+                repeatView.text = value
+            } else editText.setText(value)
+        }
+
+    fun isRepeatEnabled() = repeatModeFlag
```
BrapiCoordinatorSelby, line 94: the `isRepeatEnabled()` function is only used twice in this class, but the `repeatModeFlag` is used seven times. Consider using it every time

app/src/main/java/com/fieldbook/tracker/views/RepeatedValuesView.kt
```diff
@@ -0,0 +1,346 @@
...
+
+        mValues.clear()
+
+        mValues.addAll(values)
+
+        pager.adapter = RepeatedValuesPagerAdapter(context).apply {
+
+            submitItems(values)
+
+        }
```
BrapiCoordinatorSelby, line 231: Call `submitList()` here instead of manually setting the `pager.adapter`. It will help keep the code in sync in the long run

app/src/main/java/com/fieldbook/tracker/views/RepeatedValuesView.kt
```diff
@@ -0,0 +1,346 @@
...
+        if (pager.currentItem != (pager.adapter?.count ?: 1) - 1) {
+
+            swapRightButtonVisibility(!newButtonVisible)
+
+        } else {
+
+            swapRightButtonVisibility(newButtonVisible)
+
+        }
+
```
BrapiCoordinatorSelby, line 330: I'm pretty sure this whole if can be simplified to
```
swapRightButtonVisibility(pager.currentItem == (pager.adapter?.count ?: 1) - 1)
```
`newButtonVisible` is an unnecessary variable that is always false.

with that simplification, perhaps all the code from `swapRightButtonVisibility()` can be brought into this function

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
chaneylc: I didn't change this for now, but it is a good suggestion. This solution doesn't match exactly with how the resources are named, and this would cause a runtime error if those names are changed vs a compile time error for the current solution.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
chaneylc: Yeah, we are getting there. My plan is to separate collect activity into different files, so this will eventually happen.

app/src/main/java/com/fieldbook/tracker/database/DataHelper.java
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/traits/BaseTraitLayout.java
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/traits/CategoricalTraitLayout.java
chaneylc: I'll make an issue to update how formats are handled.

app/src/main/java/com/fieldbook/tracker/traits/LabelPrintTraitLayout.java
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/views/CollectInputView.kt
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/views/RepeatedValuesView.kt
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/views/RepeatedValuesView.kt
chaneylc: Cleaned in new commit.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
chaneylc: Cleaned in new commit.

BrapiCoordinatorSelby (approved):
ready to go imho

trife:
This is so cool!!!! Minor notes below, I've also sent a copy to Shahana for testing, but this is looking really really good.

- Add ok button to features disabled dialog
- Possible to add a leftscroll animation when adding a measure?
- Totally outside of scope, but can we add a listview dialog when the number in the corner is tapped that lists all current values with a delete button? Similar in style to the sort rows
- How is the value for table export selected? I assumed it would be the first or last value but it was the second to last when I tested.

chaneylc:
@trife Table export always exports the latest observation for me.

trife:
- I don't think blank values should be saved as observations or counted toward the total of observations. This is also confusing on file export (and probably complicates brapi things) since the blank is carried through in the export
- When deleting from the new view, the value doesn't update to reflect the current number of measures
- Rename new view `Observations`, rename `OK` to `Delete`, add neutral `Select All`

trife (approved):

PR #513, opened 2022-10-27
BrAPI Observation Syncing

zrm22:
# Description

This Pull Request will allow the user to pull down existing Observations for a BrAPI study.  This is not done automatically, but rather through a Dialog that is completely optional.  This allows for users to update fields on devices depending on what was scored on different devices

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

I have tested this for test fields stored in Breedbase pulling down ~150 observations for both BrAPI v1 and v2.  I have also tested the pagination by setting pageSize to 10 and pulling down the same study.

**Test Configuration**:
* Motorola G series phones from 2020-2022
* SDK: Recent versions of Android, unknown exactly.

# Checklist:

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
@@ -251,7 +251,7 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
     private void loadObservations() {
         System.out.println("Study DBId: "+study.getStudyDbId());
 
-        List<String> observationIds = new ArrayList<String>();
+        List<String> observationVariableDbIds = new ArrayList<String>();
 
         //Trying to get the traits as well:
         brAPIService.getTraits(study.getStudyDbId(), new Function<BrapiStudyDetails, Void>() {
@@ -260,7 +260,7 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
                 for(TraitObject obj : input.getTraits()) {
                     System.out.println("Trait:"+obj.getTrait());
                     System.out.println("ObsIds: "+obj.getExternalDbId());
-                    observationIds.add(obj.getExternalDbId());
+                    observationVariableDbIds.add(obj.getExternalDbId());
                 }
 
                 return null;
@@ -272,8 +272,8 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
             }
         });
 
-        System.out.println("obsIds Size:"+observationIds.size());
-        brAPIService.getObservations(study.getStudyDbId(), observationIds, paginationManager, new Function<List<Observation>, Void>() {
+        System.out.println("obsIds Size:"+observationVariableDbIds.size());
+        brAPIService.getObservations(study.getStudyDbId(), observationVariableDbIds, paginationManager, new Function<List<Observation>, Void>() {
             @Override
             public Void apply(List<Observation> input) {
                 study.setObservations(input);
--- a/app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
@@ -24,12 +24,16 @@ import androidx.arch.core.util.Function;
 
 import com.fieldbook.tracker.brapi.model.BrapiObservationLevel;
 import com.fieldbook.tracker.brapi.model.BrapiStudyDetails;
+import com.fieldbook.tracker.brapi.model.Observation;
 import com.fieldbook.tracker.brapi.service.BrAPIService;
 import com.fieldbook.tracker.brapi.service.BrAPIServiceFactory;
 import com.fieldbook.tracker.R;
+import com.fieldbook.tracker.brapi.service.BrapiPaginationManager;
+import com.fieldbook.tracker.objects.TraitObject;
 import com.fieldbook.tracker.utilities.DialogUtils;
 
 import java.util.ArrayList;
+import java.util.List;
 import java.util.Locale;
 
 public class BrapiLoadDialog extends Dialog implements android.view.View.OnClickListener {
@@ -42,6 +46,8 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
     private Boolean studyLoadStatus = false;
     private Boolean plotLoadStatus = false;
     private Boolean traitLoadStatus = false;
+    private BrapiPaginationManager paginationManager;
+
     // Creates a new thread to do importing
     private Runnable importRunnable = new Runnable() {
         public void run() {
@@ -59,6 +65,10 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
 
     public void setSelectedStudy(BrapiStudyDetails selectedStudy) {
         this.study = selectedStudy;
+    }
+
+    public void setPaginationManager(BrapiPaginationManager paginationManager) {
+        this.paginationManager = paginationManager;
     }
 
     @Override
@@ -82,6 +92,7 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
         studyDetails = new BrapiStudyDetails();
         updateNumPlotsLabel();
         buildStudyDetails();
+        //        loadObservations(); //Uncomment this if you want the app to sync the observations when field is loaded.
         loadStudy();
     }
 
@@ -227,6 +238,65 @@ public class BrapiLoadDialog extends Dialog implements android.view.View.OnClick
                         }
                     }
                 });
+                return null;
+            }
+        });
+    }
+
+    /**
+     * Function to load the observations.
+     *
+     * TODO: Decide if we want to delete.  It is likely a better approach to use the BrapiSyncObsDialog as it gives the user more control
+     */
+    private void loadObservations() {
+        System.out.println("Study DBId: "+study.getStudyDbId());
+
+        List<String> observationIds = new ArrayList<String>();
+
+        //Trying to get the traits as well:
+        brAPIService.getTraits(study.getStudyDbId(), new Function<BrapiStudyDetails, Void>() {
+            @Override
+            public Void apply(BrapiStudyDetails input) {
+                for(TraitObject obj : input.getTraits()) {
+                    System.out.println("Trait:"+obj.getTrait());
+                    System.out.println("ObsIds: "+obj.getExternalDbId());
+                    observationIds.add(obj.getExternalDbId());
+                }
+
+                return null;
+            }
+        }, new Function<Integer, Void>() {
+            @Override
+            public Void apply(Integer input) {
+                return null;
+            }
+        });
+
+        System.out.println("obsIds Size:"+observationIds.size());
+        brAPIService.getObservations(study.getStudyDbId(), observationIds, paginationManager, new Function<List<Observation>, Void>() {
+            @Override
+            public Void apply(List<Observation> input) {
+                study.setObservations(input);
+                BrapiStudyDetails.merge(studyDetails, study);
+                System.out.println("StudyId: " + study.getStudyDbId());
+                System.out.println("StudyName: " + study.getStudyName());
+                for(Observation obs : input) {
+
+                    System.out.println("***************************");
+                    System.out.println("StudyId: "+obs.getStudyId());
+                    System.out.println("ObsId: "+obs.getDbId());
+                    System.out.println("UnitDbId: "+obs.getUnitDbId());
+
+                    System.out.println("VariableDbId: "+obs.getVariableDbId());
+                    System.out.println("VariableName: "+obs.getVariableName());
+                    System.out.println("Value: "+obs.getValue());
+                }
+                return null;
+            }
+        }, new Function<Integer, Void>() {
+            @Override
+            public Void apply(Integer input) {
+                System.out.println("Stopped:");
                 return null;
             }
         });
--- a/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
@@ -1110,15 +1110,6 @@ public class BrAPIServiceV1 extends AbstractBrAPIService implements BrAPIService
 
                 System.out.println("Size of study details: "+studyDetails.getValues().size());
 
-                //Need to make sure the plot values are sorted before we try to save observations.
-                List<List<String>> sortedPlotValues = studyDetails.getValues()
-                        .stream()
-                        .distinct()
-                        .sorted((plot1, plot2) -> Integer.parseInt(plot1.get(plotId)) - Integer.parseInt(plot2.get(plotId)))
-                        .collect(Collectors.toList());
-
-                System.out.println("Size of study details after sort: "+sortedPlotValues.size());
-
                 for (List<String> dataRow : studyDetails.getValues()) {
                     dataHelper.createFieldData(expId, studyDetails.getAttributes(), dataRow);
                 }
--- a/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
@@ -644,16 +644,24 @@ public class BrAPIServiceV1 extends AbstractBrAPIService implements BrAPIService
             BrapiV1ApiCallBack<ObservationsResponse> callback = new BrapiV1ApiCallBack<ObservationsResponse>() {
                 @Override
                 public void onSuccess(ObservationsResponse response, int i, Map<String, List<String>> map) {
-                    // Cancel processing if the page that was processed is not the page
-                    // that we are currently on. For Example: User taps "Next Page" before brapi call returns data
-                    if (initPage.equals(paginationManager.getPage())) {
-                        updatePageInfo(paginationManager, response.getMetadata());
-                        // Result contains a list of observation variables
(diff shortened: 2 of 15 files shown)


zrm22:
@trife @chaneylc @BrapiCoordinatorSelby One thing that I could use some guidance on.  Right now I have code in BrapiLoadDialog which will pull down the observations when the field is initially loaded, but I turned it off in favor of having the user specifically requesting for the observations to be loaded in through the BrapiSyncObsDialog.  

My question is would anyone see a use for having it pull down the observations automatically when the field is loaded in.  I can see both sides where it is less user required interaction to have it auto-load, but the user might not actually want it(and it would take the field loading process more time).

Any thoughts?  If its not useful I can clean up BrapiLoadDialog a lot.

trife:
@zrm22 I think the best pattern would be to prompt the user to sync when the field is selected. If it is possible to determine if there are unsynced observations in the background prior then this notification could be dynamically displayed to the user based on the result of that check.

trife:
Is it possible to query/filter obs or studies to get a lastEdited single timestamp and compare that instead? If we can generate an efficient query to check for new/edited obs without having to pull them all, then the user-facing popup is definitely the way to go to ensure that they know that 1) they're out of sync and 2) there's an option to get back in sync.

chaneylc:
There probably is, but I think anytime we utilize network it could be considered an expensive call. So I'd say that should require an explicit user action, like having them press the sync option.

BrapiCoordinatorSelby:
tl;dr Just a manual sync option is probably good enough for _most_ users _most_ of the time
I figured out my opinion as I was writing this. I apologize for the length, its basically my whole thought process vomited out. 


Just trying to wrap my head around the possible workflows here

1. Manual Sync Option
Open Fields list -> Tap options (three dots) on existing field -> Tap Sync -> Pop-up confirm Sync -> Observations downloaded
2. Sync on Field Select
Open Fields list -> Select existing field (radio button) -> Pop-up trait import warning -> Pop-up Confirm Sync (optional based on preferences) -> Observations downloaded
3. Auto Sync on Field Download
Import Field w/ BrAPI -> Select Field -> "Preview Field" -> Pop-up field data preview -> Observations downloaded w/ study data -> "Save"
4. Manual Sync on Field Download
Import Field w/ BrAPI -> Select Field -> "Preview Field" -> Pop-up field data preview -> study data downloaded w/o obs -> "Sync Observations?" -> Pop-up Confirm Sync (optional based on preferences) -> download observations -> "Save"

Are these the basic workflows we are talking about here? Any that I'm missing or over simplified?

After writing out this list, I think I agree with @trife , workflow number two makes a lot of sense and I don't think we need to worry about syncing on the initial study download. I also agree with @chaneylc, we don't want to make an extra HTTP call every time the user taps a new field, that will be very slow. I would also have a preference setting that allows something like "Never Sync Observations", "Always Sync Observations", or "Ask Me Every Time", to control the pop-up. "Always Sync Observations" would always attempt to sync observations when a new field is selected, which could take time, so some additional Toast or loading wheel might be a good idea to give the user some feedback.

In conclusion, workflow number one is important to keep the manual option always available, This is priority number one. 
Workflow two is optional, it might be useful to some people, but it might not be that much more useful than a manual sync. If implemented it can be controlled by a preference to allow users to decide on a flow that works best for them. 
Workflow two also necessarily occurs after new Field download, so it makes workflows three and four redundant. 
To answer the original question @zrm22, In my opinion, yes you can clean up the BrapiLoadDialog code. 

Thank you for coming to my TedTalk 😝

app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
```diff
@@ -232,6 +243,65 @@ public void run() {
...
+        List<String> observationIds = new ArrayList<String>();
+
+        //Trying to get the traits as well:
+        brAPIService.getTraits(study.getStudyDbId(), new Function<BrapiStudyDetails, Void>() {
+            @Override
+            public Void apply(BrapiStudyDetails input) {
+                for(TraitObject obj : input.getTraits()) {
+                    System.out.println("Trait:"+obj.getTrait());
+                    System.out.println("ObsIds: "+obj.getExternalDbId());
+                    observationIds.add(obj.getExternalDbId());
```
BrapiCoordinatorSelby, line 263: I know this whole method might get removed, but this looks wrong, and other code should be checked for the same mistake.

FB Traits are BrAPI Variables, and the FB `TraitObject.ExternalDbId` is mapped to the BrAPI `Variable.observationVariableDbId`

Adding variable ids to a list of `observationIds` doesn't make sense

app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
```diff
@@ -1023,6 +1106,19 @@ public BrapiControllerResponse saveStudyDetails(BrapiStudyDetails studyDetails,
...
+                System.out.println("Size of study details: "+studyDetails.getValues().size());
+
+                //Need to make sure the plot values are sorted before we try to save observations.
+                List<List<String>> sortedPlotValues = studyDetails.getValues()
+                        .stream()
+                        .distinct()
+                        .sorted((plot1, plot2) -> Integer.parseInt(plot1.get(plotId)) - Integer.parseInt(plot2.get(plotId)))
+                        .collect(Collectors.toList());
+
+                System.out.println("Size of study details after sort: "+sortedPlotValues.size());
```
BrapiCoordinatorSelby, line 1120: Is this necessary? It looks like `sortedPlotValues` is only used for logging

app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
BrapiCoordinatorSelby: I see my confusion now, `observationIds` should be re-named, it is a list of variable dbids passed to `getObservations()`

app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java
```diff
@@ -1015,6 +1115,20 @@ public BrapiControllerResponse saveStudyDetails(BrapiStudyDetails studyDetails,
...
+                System.out.println("Size of study details: "+studyDetails.getValues().size());
+
+                //Make sure the plot values are sorted.
+                List<List<String>> sortedPlotValues = studyDetails.getValues()
+                        .stream()
+                        .distinct()
+                        .sorted((plot1, plot2) -> Integer.parseInt(plot1.get(plotId)) - Integer.parseInt(plot2.get(plotId)))
+                        .collect(Collectors.toList());
+
+                System.out.println("Size of study details after sort: "+sortedPlotValues.size());
```
BrapiCoordinatorSelby, line 1130: See comment above in BrAPIServiceV1

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
```diff
@@ -272,6 +272,58 @@ class ObservationDao {
 
         } ?: -1
 
+        fun insertObservation(expId:Int, model: BrapiObservation): Int = withDatabase { db ->
+            //         * @param exp_id the field identifier
```
BrapiCoordinatorSelby, line 276: why did this new method need to be created? How are other observations saved when collected in the app, or uploaded from a file?

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
```diff
@@ -0,0 +1,257 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setCanceledOnTouchOutside(false)
+        requestWindowFeature(Window.FEATURE_NO_TITLE)
+        setContentView(R.layout.dialog_brapi_sync_observations)
+        brAPIService = BrAPIServiceFactory.getBrAPIService(this.context)
+        val pageSize = context.getSharedPreferences("Settings", 0)
+            .getString(GeneralKeys.BRAPI_PAGE_SIZE, "1000")!!.toInt()
+        paginationManager = BrapiPaginationManager(0, pageSize)
```
BrapiCoordinatorSelby, line 61: the constructor `BrapiPageManager(context)` should automatically load the default pageSize from preferences.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
```diff
@@ -0,0 +1,257 @@
...
+     * Function to load the observations from a specific study
+     */
+    private fun loadObservations(study: FieldObject) {
+        val brapiStudyDbId = study.exp_alias
+        val fieldBookStudyDbId = study.exp_id
+        val brAPIService = BrAPIServiceFactory.getBrAPIService(this.context)
+
+
+        //Get the trait information first as we need to know that to associate an observation with a plot:
+        brAPIService!!.getTraits(brapiStudyDbId,
```
BrapiCoordinatorSelby, line 103: Why are you getting a new list of traits from the server? Downloading the study should have downloaded the traits already, and you have them in the FB DB. Pulling the traits from the FB DB is probably a good idea anyway, so if the user deletes a trait in app, or downloads additional ones, then you only need to sync observations for the traits actually being collected.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
```diff
@@ -0,0 +1,257 @@
...
+            { input ->
+                val observationIds: MutableList<String> = ArrayList()
+
+                for (obj in input.traits) {
+                    println("Trait:" + obj.trait)
+                    println("ObsIds: " + obj.externalDbId)
+                    observationIds.add(obj.externalDbId)
```
BrapiCoordinatorSelby, line 110: Same as comment above, `observationIds` is a confusing name and should be renamed to something like `traitIds` or `variableDbIds`

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
```diff
@@ -0,0 +1,257 @@
...
+        val mHandler = Handler()
+        mHandler.post(importRunnable)
+    }
+}
+// Mimics the class used in the csv field importer to run the saving
+// task in a different thread from the UI thread so the app doesn't freeze up.
+internal class ImportRunnableTask(val context: Context, val studyObservations: StudyObservations) : AsyncTask<Int, Int, Int>() {
+
+    var dialog: ProgressDialog? = null
+    var brapiControllerResponse: BrapiControllerResponse? = null
```
BrapiCoordinatorSelby, line 184: `brapiControllerResponse` is not used in this class, delete?

BrapiCoordinatorSelby (changes requested):
Most of the code looks good. I have a couple comments/questions to be addressed, and the workflow discussion needs to be resolved

app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
zrm22: Yes, this can be removed, I will remove this in both v1 and v2.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
zrm22: The reason why we load them here is because otherwise there is no guarantee that the trait which is on the server was loaded in when the study was loaded.  In our use case we sometimes need to add traits after the fields have already been scored for other traits.  We would then either need to reload the field or reload the traits and then do the observation sync.  This way it pulls everything associated with that study with a single reload.

We also rarely delete a trait from the app, just mark it as inactive.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
zrm22: Yeah that is wrong.  I am renaming to observationVariableDbIds to match BrAPI spec.

app/src/main/java/com/fieldbook/tracker/brapi/BrapiLoadDialog.java
zrm22: Agreed, the name is wrong.  Changing it to observationVariableDbIds to more closely match BrAPI.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
zrm22: Correct, will delete in next commit.

trife:
@zrm22 For this PR I'm satisfied with the manual syncing. We can definitely handle the optimizing the user-facing side later. Related to issue #514

zrm22:
Re: Manual Syncing - Ok I think for now then I will leave in the code in case we need it in the future.  It likely can be further refracted as well at that time so there is no code duplication.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
zrm22: Part of the reason is that this new method will check to see if there is already an observation in the DB with that id.  That being said, I should be able to convert the BrapiObserervation into a ObservationModel and just add the check to the old method.  It does look like that is not being used though currently either. 

 I can track down what the current method call is and see if we can use it here(unless @chaneylc or @trife can find it faster).

zrm22: It uses the insertUserTraits method in ObservationDao.  I think this could work as when the normal CollectActivity needs to update it will delete the observation(called Trait in the code) first and then try to load a new piece of information.  I can still put in my check to make sure we are not overwriting anything as in the normal use case there should be nothing there.

This will likely need to change once the multi-observations per plot stuff is worked out as well.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
BrapiCoordinatorSelby: I guess this is a question of functionality. I could see a use case where I have a small set of traits i'm interested in collecting, but by syncing, suddenly I have a large number of extra traits in loaded in FB I wasn't expecting. In my use case, I would have to delete traits after syncing, in your use case you would have to add traits before syncing. I guess it comes down to which use case is more likely or happens most often, and I have no idea. @trife might have some insight here. 

Or we keep it the way you have it and see if anyone complains lol

zrm22: That makes sense.  We also do see that as the season progresses.  It's not as big an issue though as the later traits we score are 1 day and done type ones which we would not sync observations as its the only day we would need to score.

Maybe what would be more useful long term is that when you click Sync Observations it gets the traits for a given study from BrAPI -> Loads a new Dialog with check boxes for each trait -> Click Load Button -> BrAPI pulls down only selected traits for that study.

This would allow the user to select which traits are needing to be synced without relying on them to load in through the Trait Dialogs.  

This would take me a bit of time to get implemented which I honestly don't have for a month or two, but I think it solves both use cases fairly elegantly.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
BrapiCoordinatorSelby: oooooo I see. I tried to cast `Context` to `Activity` in the constructor and it did not like that...
if you were feeling really generous you could modify the constructor to check `if (context instanceof Activity)` and wrap it around lines 26-30. Or just leave it and use the other constructor, fine by me either way.

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
BrapiCoordinatorSelby: I don't want to make it too complicated for you at this point, I'm just trying to avoid too many different methods accessing the DB in slightly different ways. That is how you get a corrupt DB when those methods fall out of sync. Perhaps make a new issue to address this later if it is too complex to map to one of the existing functions.

app/src/main/java/com/fieldbook/tracker/dialogs/BrapiSyncObsDialog.kt
zrm22: I will try that after I demo.  I'd rather not break it 45 minutes before I need to show it off :smile:

app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt
zrm22: Yeah I agree, use the same methods as much as possible.  

I would like to take a shot at fixing it for this PR.  If it is too involved a new issue will be created.  I think it should be an easy fix, just need to make sure I map the variables in the BrAPI object to the right variables in the method.  Not hard, just need to be careful and make sure its right.

BrapiCoordinatorSelby (approved):

trife:
@zrm22 Would you willing to make issues for any of the conversations that will require additional development efforts? Going to merge this and additional changes will be part of the winter cleanup.

zrm22:
@trife Yeah I can go through and make issues for the additional development efforts.  I will assign the ones I can do to myself. I might not have a chance to work until the beginning of the year, but I should be able to get some if not all of it ready before field season.  Thanks again for reviewing and accepting the code!

PR #515, opened 2022-10-28
Bug 512 added ability to import obs. vars. that already have FB formats defined

### LINKED ISSUE #512: [BUG] - Downloading Traits via BrAPI does not allow for all data types
**Describe the bug**
As of version 5.2.5, all traits downloaded via brapi will be `categorical`, `date`, `numeric`, or `text`. If a data type is unrecognized, it will default to `text`. This means users need to manual change the data types of traits downloaded via brapi. 

**Expected behavior**
Traits downloaded via BrAPI should be stored with correct data types and should be usable right away.

**Additional context**
Code needs to be modified in [BrAPIServiceV2.java line 939 convertBrAPIDataType()](https://github.com/PhenoApps/Field-Book/blob/master/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java#L939)

Related to https://github.com/plantbreeding/BrAPI/issues/476

chaneylc:
Added more trait format options when importing obs. vars. from brapi now the user can have obs. vars. with field book trait formats, instead of having to edit them after import.

# Description

Updated convert brapi data type function in v1 and v2 to allow for fieldbook trait formats to be used.

Fixes #512

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV1.java
@@ -942,18 +942,52 @@ public class BrAPIServiceV1 extends AbstractBrAPIService implements BrAPIService
         return sb.toString();
     }
 
+    /**
+     * @param dataType
+     * @return
+     */
     private String convertBrAPIDataType(String dataType) {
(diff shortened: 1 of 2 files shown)


BrapiCoordinatorSelby (approved):
Exactly the idea I had in mind 👍

PR #522, opened 2022-11-04
Updated sort order dialog for field editor

### LINKED ISSUE #350: Custom adapter for Field Sorting rows

chaneylc:
# Description

Basic recycler view / adapter that lets you drag and drop as well.

Fixes #350

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -100,9 +100,8 @@ import com.google.zxing.integration.android.IntentResult;
 
 import org.jetbrains.annotations.NotNull;
 import org.phenoapps.interfaces.usb.camera.UsbCameraInterface;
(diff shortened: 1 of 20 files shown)


trife:
- Issue with recyclerview- deleting a row adds a dummy row at the end (that I now cannot replicate)
- Less vertical white space/buffer/padding/etc.
- Either use the row icons as the drag/drop target (instead of long press) or remove
- Do we have any way of reverting back to original order? If so add a fourth button "Original order"
- Don't autofill with an the first field. There's no guarantee that the imported file was sorted by the first row of the field.

Overall this is a great layout, I think @timparsons will appreciate it

Screenshot_20221104-170054

trife:
- Change scrollview height to wrap list
- Scrollview has some weird behavior. If I load the branch, add three attributes, and delete one, the bottom most attribute is copied and pasted so there are still three attributes listed. Not sure if this is a visual bug or what
- Revert should go back to the original imported order- unrelated to the sort dialog (if possible)

Issue #317, opened 2021-08-06
[REQUEST] Summary data selection


ramierop:
Request is not a problem, just a possible improvement.

What I am suggesting is to be able to select what is shown in the Summary icon. 
![Screenshot_summary](https://user-images.githubusercontent.com/70235332/128537689-86dff3b9-6e00-4323-9800-fc7162b810a5.png)

As a program we are trying to get more data driven and be able to make on the fly decisions using Fieldbook and Sweetpotatobase. In an attempt to be able to make those decisions from a single tablet we are downloading information from SPB, importing that dataset to Fieldbook, and making field ratings and selections. We download traits that are helpful for field selection and can see that information in the Summary icon. It can get very busy, and I'm not sure it needs all the information that is included with the imported dataset.

![Screenshot_standardview](https://user-images.githubusercontent.com/70235332/128537756-1a597ee2-9db5-4983-adf6-5f9d1e27dc4d.png)

**Describe the solution you'd like**
The ability to select what is seen in the Summary icon.

**Describe alternatives you've considered**
We can add more databars to the Collect screen, but after that I'm unable to display the traits that I want to collect. 
![Screenshot_infobars](https://user-images.githubusercontent.com/70235332/128537829-ab698009-5d9b-4584-9929-821a29ed48fe.png)

**Additional context**

trife:
@ramierop This is a great idea. I don't have a timeline on getting it implemented, but will keep you updated via this issue.

ramierop:
@trife Nice. Thanks for the update. Looking forward to see what y'all come up with.

trife:
@ramierop This feature will be in v5.4 which will go into beta in December

ramierop:
Hi Trevor,

Thanks for the update. I will be happy to check it out and see how it
works.

Hope you're doing well.

Russell

PR #528, opened 2022-11-09
Summary update

### LINKED ISSUE #71: Link summary aspects with app features
Add ability to navigate to specific traits/data points from within summary.

chaneylc:
# Description

Added ability to filter plot attributes / traits that are viewable in the summary.

Items can be clicked to navigate to that trait.

Next/Previous button to move entries.

I did't add swipe to move entries because this was clashing with the scroll view and it didn't seem right. Also it would be a hidden feature which we typically haven't done in the past, replaced with buttons.

Fixes #71 #317

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -131,7 +131,8 @@ import kotlin.Pair;
 
 @SuppressLint("ClickableViewAccessibility")
 public class CollectActivity extends AppCompatActivity
-        implements SensorEventListener, GPSTracker.GPSTrackerListener, UsbCameraInterface {
+        implements SensorEventListener, GPSTracker.GPSTrackerListener,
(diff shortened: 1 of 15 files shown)


trife:
- Disable cvET when summary is open (when switching between plots on text trait the keyboard is displayed on top of the summary)
- Replace text buttons with black entry nav arrows
- Change edit title to "Select summary labels"
- Change toggle all behavior: only toggle selection, don't close the selection window
- Either: 1) save selected summary labels on a per field basis or 2) when switching fields, if length(summaryLabels)==0, reset current label selection
- Use materialdesignicon edit icon

trife:
- Replace next/previous bottom buttons with the entry arrows instead of on the individual trait values.
- Use accentColor for the arrows on the individual traits

trife:
- Revert back to the entire trait textview block (including arrows) navigating to the plot/trait. The design is good, just want to have more actionable 
- Add round button background to bottom navigation buttons and increase size by 25%

trife:
@chaneylc This looks great!! Go ahead and resolve conflicts/merge with develop.

Issue #445, opened 2022-05-18
BrAPI deeplink / Android 12 link issues


trife:
1. When returning from brapi auth activity, multiple versions of Field Book are displayed as options to open. This is new behavior, I think with the addition of `export=true` in certain manifest activities (required by Android 12)
2. Going to https://phenoapps.org/field-book on mobile should take the user to the app (or app store if not installed). It looks like this behavior [was changed](https://doordash.engineering/2022/01/25/your-deep-links-might-be-broken-web-intents-and-android-12/) in Android 12 as well.

trife:
If this requires a ton of work, move to v5.6

PR #577, opened 2023-05-07
OAuth: Support authorization code flow

mobreza:
# Description

[OAuth 2.1](https://oauth.net/2.1/) does not include `response_type=token` anymore, it is omitted from this specification.

This MR adds support for authorization code flow `response_type=code` and updates the relevant dependency from version 0.9.0 to (current) 0.11.1.

## Type of change

Added a new option under "OIDC Flow" called "OAuth2 Authorization Code". 

When selected, it uses `redirect_url=fieldbook://app/auth` and `response_type=code`.

`redirect_url` using custom scheme "fieldbook://.." apparently work best across different Androids, the `https://` type redirects commonly result in [missing intent data](https://github.com/openid/AppAuth-Android/issues?q=is%3Aissue+intent+null).

The implementation follows the README from https://github.com/openid/AppAuth-Android.

- [x] This change requires a documentation update

# How Has This Been Tested?

Tested only against our OAuth 2.1 server (i.e. staging.ggceapi.genesys-pgr.org) using the embedded "fieldbook" client ID.

Canceling the auth request works, but since the original code did not specify an intent for that, I did not add it.

# Checklist:


mobreza:
Related to #445

app/src/main/AndroidManifest.xml
```diff
@@ -148,7 +148,8 @@
             android:configChanges="orientation|keyboardHidden|screenSize|locale"
             android:launchMode="singleTask"
             android:screenOrientation="portrait"
-            android:exported="true">
+            android:exported="true"
+            tools:node="replace">
```
mobreza, line 152: https://github.com/openid/AppAuth-Android also mentions using "replace" when dealing with intent URLs.

chaneylc (approved):

chaneylc (commented):
Thank you for the contribution, this looks good. I tested this using cassavabase and it worked (depending on device, see below), the test server shows a web-facing error when using this method but nothing crashes and the original preferences still work. I added a small change that keeps OIDC url preference visible if you have the code flow selected, also added a check to use FLAG_MUTABLE vs FLAG_IMMUTABLE depending on the device SDK, this was causing older devices to fail authorization.

After reviewing https://github.com/openid/AppAuth-Android should we try to use reverse domain notation for the scheme so instead of fieldbook://, tracker.fieldbook.com://

Also should the host attribute be more specific here? Instead of just "app" the AppAuth-Android shows the reverse package id.
`<data
      android:host="app"
      android:scheme="fieldbook"
      android:path="/auth" />
`

Related to https://github.com/PhenoApps/Field-Book/issues/445
This does fix the disambiguation when using the code flow, but still shows with implicit flow.

Tested on Pixel4a, Pixel6 and Pixel3 with varying results:
Pixel4a code flow works but implicit now fails with these logs:
W/AppAuth: State returned in authorization response (null) does not match state from request (4W7A0Z01li-ANwrcprTMjA) - discarding response

E/BrAPI: Error starting BrAPI auth
    AuthorizationException: {"type":0,"code":9,"errorDescription":"Response state param did not match request state"}

Pixel3: implicit works but code does not work
Pixel6: both work

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiAuthActivity.java
```diff
@@ -256,40 +277,57 @@ public void checkBrapiAuth_OLD(Uri data) {
...
-
             authError(null);
-            return;
         }
     }
 
     public void checkBrapiAuth(Uri data) {
+        AuthorizationService authService = new AuthorizationService(this);
+        AuthorizationException ex = AuthorizationException.fromIntent(getIntent());
+        AuthorizationResponse response = AuthorizationResponse.fromIntent(getIntent());
```
BrapiCoordinatorSelby, line 290: The documentation on the `.fromIntent()` method says:
> Extracts an authorization response from an intent produced by {@link #toIntent()}.
Digging a little deeper, it looks like `.toIntent()` JSON serializes the response object, and adds it to a new Intent as an extra. Then `.fromIntent()` just pulls the string out of the Intent extras bundle and deserializes it. 
I ran into trouble while testing because the serialized response was not present in the Intent, so `response` was null, so the code was never exchanged for a token. 

I think we either need to figure out how to use `.toIntent()` up near line 181, or figure out how to use `AuthorizationResponse.Builder().fromUri()` and take the response directly from the redirect_uri.

BrapiCoordinatorSelby:
@chaneylc @trife Side note and Exciting news: This not only addresses #445, it completely fixes it for me.
The act of researching and testing this PR allowed me to fix the issue I was having with the `fieldbook://` scheme and KeyCloak. I have some theories, but I don't fully understand how or why yet (main suspect is a misleading error message somewhere). The point is we don't have to rely on the `https://` for my sake anymore!!

@mobreza Code looks good over all. I would approve this PR, but I ran into an issue while testing. I left some technical notes in a comment (BrapiAuthActivity 290). I don't have time this week to dig any further or try to fix the issue I was seeing. Please take a look, let me know what you think. If you have time to look at it this week great, if not I should have some more time next week to keep digging. 

Now that I have fixed my keycloak issue, anyone should be able to recreate this issue by connecting to the brapi test server auth. I can provide test credentials if you DM me on Slack

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiAuthActivity.java
mobreza: @BrapiCoordinatorSelby the `net.openid.appauth` library uses additional Intents before calling `BrapiAuthActivity` with a response/error. 

This requires `MUTABLE` pending intent in the call to `authService.performAuthorizationRequest()` (that's the first important change in this code).

I don't develop for Android regularly, but I believe `tools:node="replace"` in the AndroidManifest helps with triggering the right Intent when using app-links (fieldbook:// or https://) and addresses #445.

Only with these two changes   `.fromIntent()` works properly.

> take the response directly from the redirect_uri

I don't think that is necessary or even wanted. Fieldbook code should rely on `fromIntent()` only and let the library handle the exchange.

mobreza:
> but I ran into an issue while testing...

@BrapiCoordinatorSelby It's unclear what issue you ran into and whether you resolved it with properly configuring Keycloak or by changing the authentication code in Fieldbook.

I had success using `fieldbook://app/auth` redirect, but not with `fieldbook:/auth` (with host or even path unspecified as originally registered in https://github.com/PhenoApps/Field-Book/pull/577/files#diff-7fa6aef292187a049f7a4d6060d8df3ba212d838789c78940bd363344b1c38cdR170-R174).

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiAuthActivity.java
BrapiCoordinatorSelby: I just double checked, I'm running on build version 33, so the `MUTABLE` flag is being set (line 187). And the manifest still has `tools:node="replace"` set properly. I'm running the code from this PR/branch without edits.

But I'm still getting a `null` response and ex coming from `.fromIntent()`. I will try to dig more next week to find a root cause. @mobreza if you have a working auth server that supports the auth code flow, and I can use it, that would be helpful. I want to compare responses while debugging the FB code

BrapiCoordinatorSelby:
Ok I found the problem I was having, it has to do with the Manifest and the back stack. I can explain further if anyone cares about the details. I have committed my changes to this branch so ya'll can play with it as well.
I did some testing and got some interesting results we probably still need to figure out. For some reason Android 9 (API 28) doesn't like the code flow (no code coming back in the redirect URI), and Android 5.1 (API 22) doesn't like the Breedbase login page (tapping the login button just reloads the page instead of redirecting back to the app). My phone with the latest android seems to be working correctly across the board now. 

@mobreza @chaneylc could one of you try to recreate my failed results. Perhaps its just my virtual devices being unstable? I don't know how backward compatible the latest AppAuth library is meant to be? 
We should also maybe try it against a BMS server and maybe a BIMS server, I don't have credentials for either. They should be roughly the same as Breedbase in terms of auth services, but you never know.

Testing results

| | | Galaxy S22 API 33 (physical)       | Pixel 6 Pro API 28 (Virtual)  | Pixel 6 API 22 (Virtual) |
|---|---|---|---|---|
| genesys-pgr.org       | implicit flow  | --| --| --|
|                                 | code flow      | PASS| FAIL | PASS | 
| test-server.brapi.org | implicit flow  | PASS | PASS | PASS |
|                                 | code flow     | PASS | FAIL | PASS | 
| yambase.org             | implicit flow | PASS | PASS | FAIL |
|                                 | code flow     | --| --| --| 
| cassavabase.org       | implicit flow | PASS  | PASS | FAIL |
|                                 | code flow     | --| --| --|

BrapiCoordinatorSelby:
@nahuel-soldevilla @leetaei 
Do either of your teams have the bandwidth to test this new auth code in FB? If not, do you have a test server you could give us credentials to test?
There should be no change to your existing auth flow, we just want to make sure this new code doesn't break anything.
If we can confirm that this new code doesn't break anything for BMS or BIMS, then I think we can merge it.

nahuel-soldevilla:
Hi @BrapiCoordinatorSelby,

I think you can use [the bms brapi-uat](https://brapi.bms-uat-test.net/ibpworkbench/main) server (@ctovar to confirm and/or provide more info), it has the following users (if you don't remember the password you can use the reset flow):

```
User name | First Name | Last Name | Email
pselby | Peter | Selby | peter.selby@cornell.edu
trife | Trevor | Rife | trife@ksu.edu
chaney | Chaney | Courtney | chaneylc@ksu.edu
```

cc @mcrimi

clarysabel:
Hi @BrapiCoordinatorSelby, certainly we have a brapi server to test but the integration with FB is broken now for non admin users  due to some changes in our permissions schema. We are working on fixing it and I guess it should be solved by July 10th. 
Let me check if we can set an user that you can temporarily use for your tests. I'll reply back tomorrow.

clarysabel:
Hi @BrapiCoordinatorSelby 
I have modified the role for the users @nahuel-soldevilla has listed above. I hope you can run your tests without any problem caused by the schema permissions bugs we have in this minor version. 
I recommend to use `pselby` user because it is the one with access to more crops. 
Let me know if it helps and feel free to report any issue you may find! 

CC @mcrimi

BrapiCoordinatorSelby:
Test results for BMS same as Breedbase

| | | Galaxy S22 API 33 (physical)       | Pixel 6 Pro API 28 (Virtual)  | Pixel 6 API 22 (Virtual) |
|---|---|---|---|---|
| bms-uat-test.net             | implicit flow | PASS | PASS | FAIL |
|                                 | code flow     | --| --| --| 

Something about the old Android webview/custom tab doesn't like the newer auth flow. @trife @chaneylc your call if this is an acceptable bug to merge with. Maybe spawn a new issue to look at this later? I'm really not sure whats going on, and I think it will take some serious digging to figure it out completely.

@leetaei Please let us know if you'd like to be included in the test before this is merged.

BrapiCoordinatorSelby:
Well I feel a little dumb ... I just tested the main branch and got exactly the same issue with the API 22 Virtual device. So that is not an issue with this PR, its just an issue in general. Possibly not even related to Field Book at all. I don't know why I didn't think to try that earlier ... c'est la vie
I'll make a new issue, and we can ignore it until someone complains

chaneylc:
We decided to remove the old custom field book auth (which I included in the last commit)
I tested with a Pixel 3, Pixel 6, and Pixel 4a physical devices, they all worked on the test server with each flow.

For some reason I can't get bms-uat-test.net to work at all. If someone posts the base url and oidc I can test it further.

@BrapiCoordinatorSelby  Feel free to message me any logcat you have about the emulator not working or make a new issue.

trife:
@all-contributors add @mobreza for code

PR #609, opened 2023-05-19
fixes #607

### LINKED ISSUE #607: [BUG]: too many infobars breaks collect screen ui
### Description

too many infobars breaks collect screen ui

### Expected behavior

_No response_

### Field Book version

_No response_

chaneylc:
## Description

Too many info bars was causing collect screen mangling


trife:
I'm not a fan of this approach to solve the issue.

PR #613, opened 2023-05-19
Issue/586/infobar alert dialog

chaneylc:
## Description

New infobar dialog.


trife:
I think the parent view should be a constraintlayout so that the Cancel button is always visible without having to scroll down for long lists.

trife:
- [ ] Selector view should only activate when prefix is selected
- [ ] Bring back old behavior of expanding ellipsized text when pressHolding the infobar value

trife (approved):

PR #652, opened 2023-06-30
Add barcode scanner to main screen

### LINKED ISSUE #632: Add barcode scanner to main screen
Search all plots and all fields to load/navigate to the barcoded item.

chaneylc:
## Description

Adds button to config page that searches for unique obs unit id and switches to that field if needed. If one is found it automatically switches to the collect activity at that id. Fixes #632 

## Type of change

## Checklist:

## Changelog entry

```release-note
Added observation unit barcode scanner to main screen
```


trife (approved):

trife:
@chaneylc Would it make more sense to reuse the existing switchfield functions from the `StudyDao` and `FieldEditorActivity`? I think Collect should only open if there is a successful barcode match, so probably move that logic into Config.

chaneylc:
They do use the same function, DataHelper's switchField. Since its handling obs units I think it makes mores sense to be in Collect, since its using the already existing GeoNav feature for switching fields. But I see your point too and I can try to wrap some of that code in a utility and make that happen in Config.

Issue #659, opened 2023-07-06
[REQUEST]: Update GeoNav log to include GPS correction status


ahulsekemp:
### Description

Can you please update the GeoNav logging feature to include GPS correction status, such as that displayed currently in GNSS trait collection? (Float RTK, GPS, RTK, etc)

PR #693, opened 2023-07-31
Geonav shorter log

kamathprasad9:
## Description

Goenav now separately logs (```/storage/geonav_shorter/```) only the closest plant instead of logging all the plants in the field. This PR also addresses issue #659; location accuracy correction status is now being logged too.

## Type of change

## Checklist:

## Changelog entry

```Added separate geonav logging to capture only the closest plant```

### CODE REVIEW

**Comment by trife** [comment]
- [ ] Save to genoav directory instead of new directory
- [ ] Setting in geonav section to specify shorter vs longer log

File: app/src/main/java/com/fieldbook/tracker/preferences/GeoNavPreferencesFragment.kt [hunk 1]
```diff
@@ -65,6 +66,27 @@ class GeoNavPreferencesFragment : PreferenceFragmentCompat(),
             true
         }
 
+        val geoNavLog = findPreference<CheckBoxPreference>(GeneralKeys.GEONAV_LOG)
+        val geoNavLoggingMode = findPreference<ListPreference>(GeneralKeys.GEONAV_LOGGING_MODE)
+
+        // when GeoNav screen loads, check if GeoNav Log is checked
+        // if checked, only then display the logging mode
+        geoNavLoggingMode?.isVisible = geoNavLog!!.isChecked
```
chaneylc, line 74: Avoid using "!!" and instead wrap with a null check to avoid NPE, which would crash the app.

app/src/main/java/com/fieldbook/tracker/preferences/GeoNavPreferencesFragment.kt
```diff
@@ -65,6 +66,27 @@ class GeoNavPreferencesFragment : PreferenceFragmentCompat(),
...
+        // if the geonav is checked, only then display the logging mode selection tile
+        geoNavLog?.setOnPreferenceChangeListener { preference, newValue ->
+            geoNavLoggingMode?.isVisible = newValue as Boolean
+            true
+        }
+
+        // select geonav logging mode
+        geoNavLoggingMode?.setOnPreferenceChangeListener { preference, newValue ->
+            mPrefs.edit()
+                .putString(GeneralKeys.GEONAV_LOGGING_MODE, newValue as? String ?: "Shorter Log").apply()
```
chaneylc, line 85: Replace static strings such as "Shorter Log" with a string resource.

app/src/main/java/com/fieldbook/tracker/utilities/GeoNavHelper.kt
```diff
@@ -92,11 +92,15 @@ class GeoNavHelper @Inject constructor(private val controller: CollectController
             val altLength = alt.length
             alt = alt.substring(0, altLength - 1) //drop the "M"
 
-            //always log external gps updates
-            writeGeoNavLog(
-                mGeoNavLogWriter,
-                "$lat,$lng,$time,null,null,null,null,null,null,$fix,null,null,null,null\n"
-            )
+            val currentLoggingMode = mPrefs.getString(GeneralKeys.GEONAV_LOGGING_MODE, controller.getContext().getString(R.string.pref_geonav_logging_mode)) ?: controller.getContext().getString(R.string.pref_geonav_logging_mode)
```
chaneylc, line 95: This is used many times throughout your PR, consider making it a function.

app/src/main/java/com/fieldbook/tracker/utilities/GeoNavHelper.kt
```diff
@@ -282,7 +286,7 @@ class GeoNavHelper @Inject constructor(private val controller: CollectController
 
         writeGeoNavLog(
             mGeoNavLogWriter,
-            "start latitude, start longitude, UTC, end latitude, end longitude, azimuth, teslas, bearing, distance, fix, closest, unique id, primary id, secondary id\n"
+            "start latitude, start longitude, UTC, end latitude, end longitude, azimuth, teslas, bearing, distance, fix, closest, accuracy correction status, unique id, primary id, secondary id\n"
```
chaneylc, line 289: I recently added the 'fix' header which will give a string value such as GPS, RTK within the log. It seems you're adding a numerical value that represents the accuracy in meters from internal GPS (location.accuracy). This might crash when using external devices because accuracy has not been set within the Location object. Is it necessary to show the accuracy in meters vs already knowing it is a GPS fix? If it is necessary, this will need to be updated by adding accuracy to the external Location object when external devices are used.

app/src/main/java/com/fieldbook/tracker/utilities/GeoNavHelper.kt
```diff
@@ -536,8 +539,14 @@ class GeoNavHelper @Inject constructor(private val controller: CollectController
                         .replace(":".toRegex(), "-")
                         .replace("\\s".toRegex(), "_")
                     val thetaPref = mPrefs.getString(GeneralKeys.SEARCH_ANGLE, "22.5")
-                    val fileName =
-                        "log_" + interval + "_" + address + "_" + thetaPref + "_" + System.nanoTime() + ".csv"
+                    val currentLoggingMode = mPrefs.getString(GeneralKeys.GEONAV_LOGGING_MODE, R.string.pref_geonav_shorter.toString())
```
chaneylc, line 542: R.string.pref_geonav_shorter.toString() will return a string representing an integer id for a resource. Use context.getString like your other preference calls.

app/src/main/java/com/fieldbook/tracker/utilities/GeodeticUtils.kt
```diff
@@ -139,7 +143,7 @@ class GeodeticUtils {
 
                     val loggedString = IzString(startTime = start.time, uniqueId = coordinate.observation_unit_db_id, primaryId = coordinate.primary_id, secondaryId = coordinate.secondary_id,
                         startLat = start.latitude, startLng = start.longitude, endLat = location.latitude, endLng = location.longitude, azimuth = azimuth, teslas = teslas, bearing = bearing,
-                        distance = distance, closest = NOT_CLOSEST, fix = fix)
+                        distance = distance, closest = NOT_CLOSEST, fix = fix, accuracyCorrectionStatus = start.accuracy)
```
chaneylc, line 146: This is what I mentioned before, start might not have accuracy, especially if its from an external source.

app/src/main/java/com/fieldbook/tracker/utilities/GeodeticUtils.kt
```diff
@@ -174,8 +178,13 @@ class GeodeticUtils {
             //after a full run of IZ, update the last CLOSEST_UPDATE to CLOSEST_FINAL
             izLogArray.findLast { it.closest == CLOSEST_UPDATE }?.closest = CLOSEST_FINAL
 
-            //print the entire array to log
-            izLogArray.forEach { writeGeoNavLog(log, it.toString()) }
+            if(currentLoggingMode == "Shorter Log"){
+                //print only the closest plant to the log
+                izLogArray.forEach { if (it.closest == 1) writeGeoNavLog(log, it.toString()) }
```
chaneylc, line 183: When testing this the log was empty. it.closest == 1 should be it.closest == CLOSEST_FINAL, which is 2. 

closest: either 0, 1, or 2.
         *      0: not closest or within the trapezoid
         *      1: updated to closest and within trapezoid
         *      2: the closest point that IZ returns
After switching to CLOSEST_FINAL, the log was working.

chaneylc (changes requested):
Overall I think we need a brief description of what it means to be a verbose vs shorter log.

Verbose will log any location update, along with various steps in the IZ algorithm.

Shorter seems to not log location updates, and only logs the CLOSEST_FINAL obs. unit, which is the closest obs. unit to the user.

That being said, I think there should be a more meaningful naming and preference summary/description for these modes. I think verbose is fine, but shorter should be something more significant.

Verbose should be selected by default.

kamathprasad9:
> Overall I think we need a brief description of what it means to be a verbose vs shorter log.
> That being said, I think there should be a more meaningful naming and preference summary/description for these modes. I think verbose is fine, but shorter should be something more significant.

How does this sound?:
- Verbose logs: Logs any location updates along with algorithm steps
- Closest plant logs: Logs only the closest plant to the user at a given time based on the algorithm used

chaneylc:
> > Overall I think we need a brief description of what it means to be a verbose vs shorter log.
> > That being said, I think there should be a more meaningful naming and preference summary/description for these modes. I think verbose is fine, but shorter should be something more significant.
> 
> How does this sound?:
> 
>     * Verbose logs: Logs any location updates along with algorithm steps
> 
>     * Closest plant logs: Logs only the closest plant to the user at a given time based on the algorithm used

That's good for now, just replace plant with observation unit to generalize it.

chaneylc (approved):

PR #714, opened 2023-08-18
run release action once daily, only build if changes to app/ since last release

### LINKED ISSUE #700: GitHub Actions Release Changes
Instead of release on merge, if changes to /app, release new build once a day at 4pm ET. If the action has been run in the last 12 hours, skip.

bellerbrock:
## Description

Alters git release workflow to run once a day, and only release if there have been changes to files in the app dir since the last release.
closes #700

## Type of change

## Checklist:

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -40,7 +40,7 @@ jobs:
   build-and-release:
     runs-on: ubuntu-latest
     needs: check-app-changes
-    if: ${{ needs.check-app-changes.outputs.app_changed == 'true' }} || ${{ github.event_name == 'workflow_dispatch' }}
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'schedule' && needs.check-app-changes.outputs.app_changed == 'true') }}
     steps:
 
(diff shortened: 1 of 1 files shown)


trife:
@bellerbrock Looking at the sandbox, it doesn't look like the output from this step is being parsed correctly since it ran even when reporting 'no changes detected.'

bellerbrock:
@trife I see what you're saying, strange. I ran it manually a few times with the build job conditional just on the 'app_changed' output, both times when 'app_changed' was false the build job was skipped.

So for some reason ${{ github.event_name == 'workflow_dispatch' }} must be evaluating as true even for the scheduled workflow runs? Can't see it's value in the run details, but I'll add some logging and do some more testing.

Though that's assuming that we want to be able to do manual builds even if app/ hasn't changed. If that's not the case then the easy fix is to just remove that part of the if.

Also, would you still like scheduled runs to be skipped if the action has been run manually in the last 12 hours? Realizing I forgot to include that check as part of this PR.

trife (approved):

PR #729, opened 2023-09-13
BrAPI setting enhancements

### LINKED ISSUE #570: Hide BrAPI in import/export dialogs if it hasn't been set up
- Potentially just add a setting at the top of the brapi settings section to 'Enable BrAPI'

bellerbrock:
## Description

Adds checkbox to enable/disable BrAPI. Hides remaining prefs and removes BrAPI from import/export options if disabled.
Adds editable server display name to use in import/export dialogs instead of 'BrAPI'. Defaults to the server url minus scheme and subdomain.
Gives user the option to make BrAPI the default import and/or export option when a server is successfully authorized.

closes #570 and #699

## Type of change

## Checklist:

## Changelog entry

```release-note
BrAPI setting enhancements
```


trife:
@bellerbrock add a check to the server name so that if it's blank it shows "BrAPI". Are there checks for if servername returns blank?

app/src/main/java/com/fieldbook/tracker/preferences/BrapiPreferencesFragment.java
```diff
@@ -111,10 +145,12 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
         }
 
         //set saved urls, default to the test server
-        String url = prefMgr.getSharedPreferences().getString(GeneralKeys.BRAPI_BASE_URL, getString(R.string.brapi_base_url_default));
-        String oidcUrl = prefMgr.getSharedPreferences().getString(GeneralKeys.BRAPI_OIDC_URL, getString(R.string.brapi_oidc_url_default));
+        String url = sp.getString(GeneralKeys.BRAPI_BASE_URL, getString(R.string.brapi_base_url_default));
+        String displayName = sp.getString(GeneralKeys.BRAPI_DISPLAY_NAME, getString(R.string.preferences_brapi_server_test));
```
chaneylc, line 149: In TraitEditor and Config activities you do this:
`String displayName = ep.getString(GeneralKeys.BRAPI_DISPLAY_NAME, "BrAPI");`

It seems this should be consistent and all cases should be R.string.export_source_brapi, but let me know if you did this for a specific reason.

chaneylc (changes requested):
- [x] disable toolbar authentication if BrAPI is disabled

app/src/main/java/com/fieldbook/tracker/preferences/BrapiPreferencesFragment.java
bellerbrock: The idea was that when BrAPI is first enabled the url defaults to the test server, so the display name should default to the test server text. From that point forward the BRAPI_BASE_URL is always set to something, so the default can b anything, it no longer gets used. But regardless you're right it should be consistent if only for readability, I'll go ahead and unify them.

Speaking of the default display name, how do you feel about changing the `preferences_brapi_server_test` string from '**Default Test Server**' to '**BrAPI Test Server**'? I think that could help clarify things.

trife: Yeah that's fine with me

trife (approved):

PR #746, opened 2023-09-25
Logging

trife:


trife:
@bellerbrock Can you merge the two logging prefs so that 'GeoNav Log' has three options: Off, Limited, Full?

bellerbrock:
@trife they are merged now, haven't been able to test the whole functionality though without an appropriate device.

PR #765, opened 2023-10-11
Field level audio (beta)

kamathprasad9:
## Description

Added field level recording. Enable this setting in beta section.

## Type of change

## Checklist:

## Changelog entry

```Added field level audio recording feature```


kamathprasad9:
Should this feature work only if Geonav is enabled by user, or should it work irrespective of that?

kamathprasad9:
If we are recording field level audio, and if we try to record trait level audio (and vice versa), that will work properly - recording the same audio into both files. Should we let this be?
Right now, I am checking the state of field level audio before letting the trait audio take over.

chaneylc:
> If we are recording field level audio, and if we try to record trait level audio (and vice versa), that will work properly - recording the same audio into both files. Should we let this be?

From my understanding of the required functionality this should be fine.

chaneylc (changes requested):
It looks good so far, I'm going to request some minor changes to keep the code base consistent.

Your FieldAudioObject.java file has state, and I recommend calling this a helper and moving it to utilities; field book objects are typically just model files, classes that hold data. Similarly, it looks like a lot of this code is from the audio trait layout, is it possible to use this helper you've created within that trait code as well? Finally, you should use Hilt injection in the collect activity to initialize this helper, look at InfoBarHelper for an example; sometimes this isn't possible, and if it's not provide documentation as to why.

I'd also like to see more documentation around FieldAudioObject, specifically, a top-file description of the feature and how it is used.

https://developer.android.com/training/dependency-injection/hilt-android 

Otherwise this looks like an awesome feature that you were able to finish quickly, great job :)

chaneylc:
Awesome, I will review this and test it today. You will need to sanitize the saved audio file names using the FileUtils sanitizeFileName function I added in this PR: https://github.com/PhenoApps/Field-Book/pull/773/files

chaneylc (changes requested):
- [x] Update the toast string when you're playing trait audio and try to record field audio, it says you're recording but you're actually just playing audio.

Everything seems to work well, so great job on the functionality. <s>My only concern is the location of where field audio is stored. You're storing it in plot_data/field_audio, if a user names a trait field_audio, its plot data will be stored along side the field_audio files. I tested this with images and nothing as far as I can tell breaks, but I don't think we want to do this.

@trife suggestions on this? Should we make a field_audio folder within the root directory? Or are you fine with the conflicting plot_data entries?</s>

[edit] upon merging https://github.com/PhenoApps/Field-Book/pull/773 we can keep using 'field_audio' as a folder and these issues will be moot.

trife:
@chaneylc or @kamathprasad9 can you resolve the conflicts in this branch? 
Use `ic_tb_field_mic_on.xml` for the icon in `preferences_beta.xml`

trife (approved):

chaneylc (approved):

Issue #171, opened 2020-10-08
Capture exif tag data on collected images


trife:
UserComment JSON containing 

- [ ] Device model
- [ ] Serial number (or some kind of device ID)
- [ ] observationUnitID
- [ ] observationVariableId
- [ ] Location

trife:
Potentially important for GOOFBALL data collection

PR #773, opened 2023-10-16
Photo file name fixes for special characters

### LINKED ISSUE #641: [BUG]: Pipe character in ontology traits causes image to fail
### Description

When I import traits from BrAPI, they are named along the lines "SomeTrait|OntoName:000001". When I try to take an image for an imported trait, I am brought the the camera, take the picture, am shown a preview, press the check mark, but then no thumbnail appears. In the file browser, there is a directory under plot_data/FIELD/TRAIT, but no images in that directory.

When I create my own trait with a name like "SomeTrait", it works as expected. When I rename my own trait to "SomeTrait|xyz", it fails in the same way described above; so I think the pipe character may be reserved or otherwise causing this to fail.

Hardware: moto g play (2021), Android version 11

### Expected behavior

Image capture works as it should even though there are reserved character(s) in the trait name. OR - on the breedbase side, trait ontology names don't include problematic characters ( I saw a related issue having to do with colons on Windows machines here: https://github.com/solgenomics/sgn/issues/4104)

### Field Book version

5.4.6.11

### LINKED ISSUE #771: [BUG]:  Camera trait fails depending on trait name
### Description

Latest version of app. Default  trait export from our breedbase is the trait name  with "|UNH:XXXXX' ontology number . We found that photo traits don't work with the "|UNH:XXXX" part in the name. The camera opens and takes a photo, but the photo doesn't save.  When the "|UNH:XXXX" part is removed the photo saves. Only seems to be an issue for photo traits. See attached working and not working trait files.
[notworking_photo_booth.txt](https://github.com/PhenoApps/Field-Book/files/12920002/notworking_photo_booth.txt)
[working_photo_booth.txt](https://github.com/PhenoApps/Field-Book/files/12920004/working_photo_booth.txt)

### Expected behavior

_No response_

### Field Book version

_No response_

### LINKED ISSUE #813: [BUG]: Barcode reader bug
### Description

Barcode reader scans the text and moves to next entry but the value is not saved.

### Expected behavior

_No response_

### Field Book version

_No response_

chaneylc:
Bugfix for saving photo traits with trait names that have special characters.

Closes #771
Closes #641 
Closes #813 

## Description

## Type of change

- [x] `BUGFIX` (non-breaking change which fixes an issue)

## Checklist:

## Changelog entry

```release-note
Fixed bug that broke photo saving for trait names with special characters.
```


chaneylc:
After digging into this a bit more I realized that the document provider API does automatically sanitize file names when using createFile, so plot data files will automatically work. BUT, createDir does not sanitize file names, so the combination of this and using trait names which could have special characters creates these issues, because they are the names of the directories.

Using the suggestions from https://github.com/PhenoApps/Field-Book/issues/641 I have changed the implementation to make folders based on trait db ids. This change does not require any migration.

chaneylc:
- [ ] update audio traits to use trait db id
- [ ] update usb trait to not use plot ids
- [ ] test go pro trait

trife:
This isn't currently working. With my test trait, `Test:||Photo`, capturing a photo creates two directories with no images saved.

app/src/main/java/com/fieldbook/tracker/utilities/FileUtil.java
```diff
@@ -27,6 +27,11 @@ public final class FileUtil {
 
     private static final String PRIMARY_VOLUME_NAME = "primary";
 
+    //https://stackoverflow.com/questions/2679699/what-characters-allowed-in-file-names-on-android
+    public static String sanitizeFileName(String name) {
+        return name.replaceAll("[|\\?\\*<\"\\\\:>\'\";]*", "_");
```
bellerbrock: Hey @chaneylc, just aheads up that this sanitizing strategy will break Breedbase's simple image upload. It assumes photos are named [observationUnitName]_trait_number_timestamp.jpg, so extracts the observationUnitName by capturing everything before the third-to-last underscore.

chaneylc: Ah, I'm assuming breedbases's simple image upload looks like this:
1. collect images with FB
2. export bundle from FB
3. manually upload images into breedbase UI

Exporting the images via brapi should remove the necessity to parse a file name.

Breaking this requirement down: [observationUnitName]_trait_number_timestamp.jpg
1. Are brackets really required around observationUnitName?
2. Obs unit names and trait names often are imported into FB with special characters, Windows and Android file systems do not allow certain characters such as '|'. FB tracks the original name in the database but we have to sanitize these names to make user-friendly-displayable folders for plot data.
3. I noticed that a lot of *Base trait names are very complicated: browsing cassavabase I see
     abiotic stress trait|month 10|COMP:0000054 
My question is, how would you previously import this with the simple breedbase image upload? I can't create a file with this name on my Windows or Android filesystem. 

Would a simple solution for this be changing the breedbase image upload parsing to instead parse the first token after a "_" split?

trife: @bellerbrock @lukasmueller Current thought right now is that we'll sanitize but write the unabridged info to a json in the exif data for each image (#171). That should allow for straightforward parsing but will require a few updates to keep the simple image upload working.

bellerbrock: @chaneylc yep, that's the current workflow. Automating image transfers with brapi would certainly be nice, and would solve this issue.

1. no sorry for the confusion, i just added the brackets to indicate that the observationUnitName is the extraction target

Regarding 2 and 3, my understanding is that photos are traits only within Field-Book to simplify their collection, they aren't defined as formal traits in crop ontology ontologies. By Crop Ontology and Breedbase standards they are akin to nirs spectra, data that is uploaded and associated with specific observation units and then used to derive trait values after analysis. In other words you should not have to worry about naming files with complex crop ontology codes and special characters, the photo trait should be created locally and should exclude them.

A potential solution off the top of my head would be to disallow importing traits of type photo via brapi. And enforce additional restrictions on the trait name (exclude special characters and underscores) when a user creates any trait that will be used to name a file within Field-Book (photo, audio, etc).

Or sticking with the sanitizing strategy you could encode/decode special characters, backslash to escape them, or remove them entirely.  @lukasmueller may have more ideas.

bellerbrock: Ah, missed that comment Trevor. I like the idea of parsing the exif data rather than the filename, that makes a lot more sense

lukasmueller: The standard encoding of special characters in the filesystem is to pre-backslash them. This often works transparently with other software downstream

lukasmueller: but as Trevor pointed out, it doesn't work in Windows :cry:

chaneylc:
Added Exif implementation, each image is saved with:
```
/**
         * Json example output saved in TAG_USER_COMMENT EXIF attribute
         *
         * {"study": {
         *  "study_name":"field_sample",
         *  "study_alias":"field_sample",
         *  "study_unique_id_name":"plot_id",
         *  "study_primary_id_name":"row",
         *  "study_secondary_id_name":"plot",
         *  "date_import":"2017-06-15 05:32:44-0700",
         *  "count":200,"internal_id_study":1
         *  },
         *
         * "observation_unit":
         *  {"observation_unit_db_id":"13RPN00002",
         *  "primary_id":"1",
         *  "secondary_id":"2"
         *  },
         *
         *  "observation_variable":{
         *  "observation_variable_name":"Maize Height",
         *  "observation_variable_field_book_format":"photo",
         *  "default_value":"",
         *  "external_db_id":"variable1",
         *  "trait_data_source":"test-server.brapi.org",
         *  "observation_variable_details":"Test"}}
         */
```

Also fixed #803

trife:
@chaneylc I think to be all-encompassing, we should include timestamp and collector in the JSON

trife (approved):

Issue #270, opened 2021-07-09
Add indicator that data has been saved in addition to color


trife:

trife:
Italicize data that is unsaved (newly entered or edited)

Related to #414

PR #807, opened 2023-11-12
View observation metadata

### LINKED ISSUE #530: Ability to view metadata of a collected observation

kamathprasad9:
## Description

Fixes #530 

## Type of change

## Checklist:

## Changelog entry

```Check metadata of observation```

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -7,6 +7,7 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.res.ColorStateList;
 import android.content.res.Configuration;
 import android.location.Location;
 import android.net.Uri;
@@ -84,6 +85,7 @@ import com.fieldbook.tracker.views.RangeBoxView;
 import com.fieldbook.tracker.views.TraitBoxView;
 import com.getkeepsafe.taptargetview.TapTarget;
 import com.getkeepsafe.taptargetview.TapTargetSequence;
+import com.google.android.material.floatingactionbutton.FloatingActionButton;
 import com.google.zxing.integration.android.IntentIntegrator;
 import com.google.zxing.integration.android.IntentResult;
 
@@ -197,6 +199,8 @@ public class CollectActivity extends ThemedActivity
     private TraitBoxView traitBox;
     private RangeBoxView rangeBox;
     private RecyclerView infoBarRv;
+
+    private FloatingActionButton observationInfoButton;
 
     /**
      * Trait-related elements
@@ -479,6 +483,15 @@ public class CollectActivity extends ThemedActivity
         infoBarRv = findViewById(R.id.act_collect_infobar_rv);
         infoBarRv.setLayoutManager(new LinearLayoutManager(this, LinearLayoutManager.VERTICAL, false));
 
+        observationInfoButton = findViewById(R.id.observationInfoButton);
+
+        observationInfoButton.setOnClickListener(view -> {
+            ObservationModel observationModel = getCurrentObservation();
+            Map<String, Object> message = observationModel.showNonNullAttributesDialog();
+
+            showObservationMetadata(message);
+        });
+
         initCurrentVals();
 
         Log.d(TAG, "Load screen.");
@@ -662,6 +675,8 @@ public class CollectActivity extends ThemedActivity
             } else {
                 traitLayouts.deleteTraitListener(getTraitFormat());
             }
+
+            updateObservationInfoButton();
 
             triggerTts(deleteTts);
         });
@@ -2199,6 +2214,7 @@ public class CollectActivity extends ThemedActivity
 
     @Override
     public void inflateTrait(@NonNull BaseTraitLayout layout) {
+        Log.d(TAG, "inflateTrait: ");
         getTraitLayout().onExit();
         View v = LayoutInflater.from(this).inflate(layout.layoutId(), null);
         LinearLayout holder = findViewById(R.id.traitHolder);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java [hunk 1]
```diff
@@ -479,6 +483,15 @@ private void loadScreen() {
         infoBarRv = findViewById(R.id.act_collect_infobar_rv);
         infoBarRv.setLayoutManager(new LinearLayoutManager(this, LinearLayoutManager.VERTICAL, false));
 
+        observationInfoButton = findViewById(R.id.observationInfoButton);
+
+        observationInfoButton.setOnClickListener(view -> {
+            ObservationModel observationModel = getCurrentObservation();
+            Map<String, Object> message = observationModel.showNonNullAttributesDialog();
+
+            showObservationMetadata(message);
```
chaneylc, line 492: Instead of using set clickable and manually changing the tint, can you use setEnabled? This should handle the tint and disable click events.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -2411,4 +2429,50 @@ public Location getLocation() {
...
         return gps.getLocation(0, 0);
     }
+
+    // updates the state of the info button
+    @Override
+    public void updateObservationInfoButton() {
+        ObservationModel currentObservation = getCurrentObservation();
+        if(currentObservation == null){
+            // if no observation is found, disable click
+            observationInfoButton.setClickable(false);
```
chaneylc, line 2439: Instead of using set clickable and manually changing the tint, can you use setEnabled? This should handle the tint and disable click events.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -2411,4 +2429,50 @@ public Location getLocation() {
...
+
+    private void showObservationMetadata(Map<String, Object> message) {
+        AlertDialog.Builder builder = new AlertDialog.Builder(this, R.style.AppAlertDialog);
+
+        StringBuilder formattedAttributes = new StringBuilder();
+        for (Map.Entry<String, Object> entry : message.entrySet()) {
+            formattedAttributes.append(entry.getKey()).append(": ").append(entry.getValue()).append("\n");
+        }
+
+        builder.setTitle("Metadata")
```
chaneylc, line 2469: Make this a string reaource

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
kamathprasad9: I had tried doing that, but the tint won't change when that is done

kamathprasad9: Now completely hiding FAB if no observation was found

trife:
@chaneylc Since the metadata is on a per obs basis, it's going get really cluttered to add each obs metadata to the summary view. Having the dialog as a basic informational 'stub' seems like the best approach. I think both this and summary view could use some visual TLC.

chaneylc:
> @chaneylc Since the metadata is on a per obs basis, it's going get really cluttered to add each obs metadata to the summary view. Having the dialog as a basic informational 'stub' seems like the best approach. I think both this and summary view could use some visual improvements, but they should stay separate.

That makes sense, I still think its best to switch to the summary's recycler view / adapter approach instead of using a string builder.

trife:
That part I agree with. If it's possible for the current summary view styling to be wrapped into a dialog, then let's do that instead of this approach.

trife:
@kamathprasad9 This looks great!!

Since we're testing a new device with limited screen real estate, I think it makes more sense to open the dialog when the current value is pressed instead of having the FAB to ensure the FAB doesn't overlap any other UI elements. Excellent work getting this added!

kamathprasad9:
I agree that FAB can/would be overlapped by other elements.

> when the current value is pressed

But do we want this to work the same way for all the traits? Text type trait (eg. notes) when pressed brings up the keyboard to edit the text. Photo trait can have multiple images for an observation, each of which when pressed opens up the respective image

trife:
That's a really good point. I think leave the FAB for now and we can see think about what a better solution could look like.

chaneylc (changes requested):
Overall it looks good, I'm just requesting the one change for the bug I found. Otherwise, I have a couple of thoughts @trife. I don't like the metadata FAB, I think its taking too much space for our trait layouts, I think it should be a toolbar icon which is linked to a preference. Alternatively it could be an extension of the repeated values delete dialog, but of course this would have to be enabled even when repeated measures is not enabled. Also think the current dialog if chosen to proceed needs some styling.

trife:
@kamathprasad9 this is looking good and I think I know the path forward. For the dialog- It's missing a close button and the margins can be adjusted to be in line with the title.

Second- I think merging this with #270 (and eventually #519) is the way to go. The ultimate goal would be to be able to long press the observation (individual images included) and display the metadata dialog you've created.

chaneylc:
- [ ] Doesn't work for Text traits
- [ ] Multi categorical trait values only show the first category 

Otherwise it looks great!

chaneylc (changes requested):
See above comments

trife:
@chaneylc if you're good with this I can merge it before the next build. Please make sure that single clicks in the upcoming camera builds open the preview since I don't think they do in this one.

chaneylc:
@kamathprasad9 everything looks good but there is one issue with photo traits:
Previously you could single-click captured images which would display the image through a system intent, with these updates you can't do that anymore. I'll approve the PR upon that fix, thanks.

trife:
Great job @kamathprasad9 !

Issue #806, opened 2023-11-10
Update search UI


trife:
- Refer to field sort UI for reference
- Use CollectAttributeChooserDialog instead of spinner for data field selection
- Create dialog with list of operators and icons instead of spinner
- Get icons from [here](https://pictogrammers.com/library/mdi/)
- Add option to sort results

![image](https://github.com/PhenoApps/Field-Book/assets/1869534/80c879a9-4cf8-4032-9377-24a5faca9741)

Issue #503, opened 2022-09-14
[BUG] Update Search Activity to a dialog


chaneylc:
Non-breaking bug.
Search Activity should be a dialog, right now it is an activity that resembles a dialog.

trife:
Related to #161

PR #825, opened 2023-11-29
Updated search UI and changed search activity to dialog

akshat22:
## Description

- Changed search activity to a dialog #503 
- Updated search UI #806 
- Displaying selected attributes in the search results dialog #161 

## Type of change

## Checklist:

## Changelog entry

```release-note

```


trife:
@akshat22 I think this is looking pretty good. I think the big leftovers here are to clean up the logic to ensure that queries accurately return results and then to work on improving the results page. What might make sense is to custom build the results page based only on the fields used in the querybuilder. Blank values could just be a global match.

For design, it would be good to have the builder/results take a pageview approach so that the user could go back to the query builder from the results. Something similar to the new trait creator.

chaneylc (changes requested):
Great job on using existing code and utilizing good Android patterns for creating adapters to solve this. I added a fewrequested changes. It's important we update the business logic of the searching, its very outdated. Also, I'm not sure if this is an AS bug or what, but I can't checkout your latest commits, and only receive the ones from Nov 8th. So I can't test this or actually view the changes, but the PR does show the code changes so I was able to review that.

trife:
@akshat22 This is looking really great! Notes:

- [ ] Increase minimum height of results dialog by 30%
- [ ] If selected columns are duplicates of the primary/secondary, don't add them to the results dialog (e.g. row/plot from the sample data)
- [ ] Replace the custom layouts with AlertDialogs (#872)
- [ ] Increase search dialog height by 25%
- [ ] Right justify text in querybuilder edittexts

trife:
@akshat22 - few other things I noticed or think should be changed in the most recent version:

- [ ] The defined query should always be saved until the user clears it. It looks like currently the behavior is to clear on 'Close'
- [ ] If the user navigated to an entry from the results page, open the results page when they reopen search
- [ ] Move the 'Back' button on results page to the Neutral spot
- [ ] Remove the reset circle button and add a 'Clear' Neutral button to the Search dialog
- [ ] Rename 'OK' on search dialog to 'Search'
- [ ] Add 4dp padding to top/bottom of search result entries to make them easier to press
- [ ] I'm not sure what the text size is set to in the results, but you should choose one that's dynamically defined from the theme sizes
- [ ] Don't close the search dialog if there are no results
- [ ] Remove right side scroll bar from search results- not sure about other people but I always end up accidentally pressing this
- [ ] Not sure how easy this is... potentially get rid of black lines between search results and use alternating white/grey shades to demarcate every other row. If there isn't an easy way to implement, just let me know

app/src/main/java/com/fieldbook/tracker/dialogs/SearchDialog.java
```diff
@@ -0,0 +1,602 @@
...
+            }
+
+            if (isAttribute) sql += sql1 + " and " + prefix;
+            else sql += sql2 + " and " + prefix;
+
+            if (i != (dataSet.size() - 1)) sql += " INTERSECT ";
+
+        }
+        return sql;
+    }
```
chaneylc, line 478: If we are going to follow this strategy it should be refactored into a query builder class that is a dependency of data helper or the obs. unit. property dao. I think it would be a powerful addition to implement sqlite EXCEPT to go along with the INTERSECT operator.

app/src/main/java/com/fieldbook/tracker/dialogs/SearchDialog.java
```diff
@@ -0,0 +1,602 @@
...
+
+                // Save flag so the search results dialog will be opened instead of search dialog
+                openResults = true;
+
+                // When you click on an item, send the data back to the main screen
+                CollectActivity.searchUnique = data[position].unique;
+                CollectActivity.searchRange = data[position].range;
+                CollectActivity.searchPlot = data[position].plot;
+                CollectActivity.searchReload = true;
+                dialog.dismiss();
```
chaneylc, line 583: Replace this static variable usage with an interface

chaneylc (changes requested):

app/src/main/java/com/fieldbook/tracker/dialogs/SearchDialog.java
akshat22: What will be the use of EXCEPT here? Is it to replace INTERSECT like `C = A EXCEPT B; Solution = A EXCEPT C`?

trife:
This looks great!!! If you resolve the conflicts I can get it merged in after Chaney reviews.

trife (approved):

chaneylc (approved):

PR #898, opened 2024-02-23
Canon Camera Integration

chaneylc:
## Description

Created a PTPIP Canon API, tested on Powershot V10.

Here is the start to a description on how to use and what the current implementation state is like, more details to come.

Updated new trait dialog to have single "Photo" trait, when chosen, the user selects the type of camera they want to use as a parameter.

Added a generic camera trait ui implementation, but other camera traits still need to be migrated to it.

Added a preference for Canon with the following: wifi ssid name, ip address, port, debug, help.
1. SSID Name: defaults as "Canon" must be a string contained within your camera's AP's SSID, e.g if your camera shows V10_abcdef_Canon then you could use "V10_abcdef_Canon" or "Canon" or "V10", etc.
2. ip address should most likely always be 192.168.1.2, but since this is assignable in the camera ui, this must be changeable here.
3. similar to point 2 but for the port of the camera
4. debug flag logs some simple messages to logcat, and if a specific type of exception occurs then a log file is created in the storage cache (this can be improved)
5. help -> this should just navigate to a tutorial on how to use the trait / camera (right now just navigates to docs webpage)

Setting up the canon:
This implementation relies on setting the camera into an AP mode. Cameras will have different ways to do this, but generally go to the connection settings and find 'Connect to smartphone(tablet)', choose 'Add a device to connect to', choose 'Camera access point mode',  choose 'Manual connection', use all the defaults EXCEPT encryption must be None, as AES does not work on Android devices. This will start a connection dialog on the camera, once a canon trait is created on FB and navigated to, a connection will be acknowledged. When the connection is acknowledged, a prompt will show on the camera. This should only need to be done once. Later when turning the camera on/off, there should be a "Field Book" option instead of "Add a device to connect to".

When navigating to the canon trait from config activity, the user will see a connection button. The connection button is not necessary, but it will reset the network scan. If a connection is already open, the camera trait will resume the connection, even when navigating to other plots. Capturing a picture will save to the plot it was captured on, even if navigated away. If a connection has not happened yet, a system network dialog should eventually pop-up showing the AP SSID (which should contain your preference SSID string). I've noticed that Android doesn't like to update the network lists, sometimes if you don't see this dialog you have to pull down your settings menu, click the "Internet" chip-button, wait for your network to be populated in the system list, then navigate back to FB (there's some room for improvement here)

If something does not work, make sure you enable debug mode (then reproduce the problem), and send me your log cat, along with the file in storage/Android/data/com.fieldbook.tracker/cache/log.txt

Images are pulled at roughly 60FPS (max), but its likely slower than that. Images are captured when the virtual shutter or the physical shutter is pressed. Images that are downloaded are native to the camera, and not preview/thumbnails. Images take a few seconds to download.

Moving back to the config activity will disconnect from the API and the AP, which means the user must restart the AP mode and select "Field Book" on their camera. 

The current implementation finds the first available storage id. Canon's can have multiple mount points, but only the first is ever used in this implementation.

Following to-do's:

- [ ] test on other devices (right now its not guaranteed to work on other devices)
- [ ] add canon tutorial to documentation, update preferences link to the tutorial page
- [ ] improve device debugger
- [ ] refactoring
- [ ] migrate other camera traits to generic camera trait
- [ ] wrap implementation into alpha settings
- [ ] figure out what android api we want to target, some of the wifi changing code is using newer apis, which won't work on older phones
- [ ] improve canon connection ui / workflow
- [ ] test capturing photo with physical shutter then navigating
- [ ] test cameras with multiple storage mounts

Misc:
fixed a bug in a unit test that was breaking build
added geonav preferences to search indexing in preferences

## Type of change

## Checklist:

## Changelog entry

```release-note
Added support for Canon cameras
```


trife:
Broken into separate PRs

PR #902, opened 2024-03-01
Include sortBy parameter in BrAPI field import

bellerbrock:
## Description

Adds a spinner to choose sortBy parameter in BrAPI field imports. Sorting by studyDbId rather then studyName allows faster access to most recently created fields in a many-paged list result.

## Type of change

## Checklist:

## Changelog entry

```release-note
sortBy parameter added to BrAPI field import
```


trife:
I'm not sure if this is working for me. I was testing with cassavabase but the order of returned studies didn't change regardless of the choice I selected.

bellerbrock:
Try with sweetpotatobase.org - the sortBy parameter has been part of BrAPI spec, but breedbase only just added it and I think not all their sites are updated with the latest release.

But come to think of it maybe there should be a way to test if the server has sortBy implemented, and hide the option if it isn't going to work.

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiActivity.java
```diff
@@ -107,7 +114,25 @@ private void loadToolbar() {
...
+        Spinner sortBySpinner = findViewById(R.id.sortBySpinner);
+        ArrayAdapter<CharSequence> sortByAdapter = ArrayAdapter.createFromResource(this,
+                R.array.brapi_study_sort_by_options, android.R.layout.simple_spinner_item);
+        sortByAdapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        sortBySpinner.setAdapter(sortByAdapter);
+        sortBySpinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parent, View view, int position, long id) {
+                sortBy = parent.getItemAtPosition(position).toString();
+                Toast.makeText(BrapiActivity.this, "New sortBy option selected: " + sortBy, Toast.LENGTH_SHORT).show();
```
chaneylc, line 128: Your toast should use a string resource.

chaneylc (changes requested):
My only other comment is possibly renaming the sortBy spinner options to be something more readable, but I get if we want to keep those completely brapi oriented. I thought things like: Study Name, Study ID, Trial Name, ... might be better.

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiActivity.java
bellerbrock: Good point. Was only needed for debugging anyway, so just removed it

trife:
@bellerbrock go ahead  Replace sort options with user-friendly strings. I'm also not sure if sorting by DbIds makes much sense since they're usually arbitrary. Planting/creation date or a similar timestamped field would be a good addition.

bellerbrock:
Here's some [more info from the BrAPI spec](https://github.com/plantbreeding/BrAPI/blob/eed6a7a2dc4dc53d4a93743c5307fce47f43e4ac/Specification/BrAPI-Core/Studies/Studies_GET_POST.yaml#L56-L82). Users can choose what they want to sort by from those options, but no date fields in there.

Which makes some sense to me because not all studies will have associated dates. But in all cases I've seen (Breedbase,T3,BIMS) studyDbId is an incrementing number that serves as the primary key in the table that stores studies, so it will sort in the same order as a create date would.

I'll switch to user-friendly strings and convert this to a draft for now. I want to talk to Pete and make a few other refinements before merging. like including sortOrder and only providing the sort option if sorting is actually implemented by the server.

trife:
@bellerbrock this can be closed, right? It looked like this was merged with the datagrid sorting

bellerbrock:
@trife Ah, I see the confusion since they both refer to 'sort order' implementation, but no the functionality is distinct.

This PR adds sortBy and sortOrder parameters to the study GET request, while the datagrid PR added an additional attribute spinner to the BrAPI field import's preview dialog to ensure that the study is saved in Field Book with a sort order (important for cases where the obs units were being imported in random order).

bellerbrock:
Closing this, this and many more improvements are already handled by #1073

PR #945, opened 2024-03-28
Offbeat camera parent branch

chaneylc:
## Description

Parent branch for canon, gopro and usb offbeat updates

This updates the behavior of the default photo trait.

USB and GoPro traits should not be tested here, they will be updated when merged with the respective branches.

## Type of change

## Checklist:

## Changelog entry

```release-note

```


trife:
@chaneylc I think this is looking pretty good with a few minor changes

- [ ] Rename option to "Default camera app"
- [ ] Set as default
- [ ] Keep preview same aspect ratio as captured image and don't rotate thumbnails (maybe stagger it so captured thumbnails are to the left of the preview and the preview itself is part of the gallery?)

trife:
It looks like only a limited amount of exif data is being written to the files that are captured on this and the USB branch

trife:
More so related to main but fixable in this branch: metadata timestamp format is currently yyyy-mm-dd-hh-mm-ss but should be the same as other timestamps, e.g. `2024-04-24 00:12:19.622-04:00`

trife:
@chaneylc I'm planning on merging this into `field_modifications`. Any reason why we should avoid that?

Issue #890, opened 2024-02-17
Add dialog when switching languages


trife:
"Warning; Translations are crowd-sourced and may not be entirely accurate or complete."

PR #971, opened 2024-05-13
Ux enhancements - explanatory info on lock and language changes

### LINKED ISSUE #915: Add toast to notify user that input is locked
- [x] Toast when trait layout area is touched when locked
- [x] Toast describing each mode as the locked icon is toggled
- [x] Add bonk sound when interacting with UI when locked

bellerbrock:
## Description

Adds toasts when locked status is changed and toast plus error sound when locked interface is clicked.
Adds dialog when language is switched to warn that translations are crowd-sourced and may be incomplete  
Closes #915 and #890

## Type of change

## Checklist:

## Changelog entry

```release-note
Explanatory info added to lock and languages changes
```

### DIFF
--- a/app/src/main/res/layout/activity_collect.xml
+++ b/app/src/main/res/layout/activity_collect.xml
@@ -143,4 +143,16 @@
 
     </ScrollView>
 
+    <View
+        android:id="@+id/lockOverlay"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:background="@android:color/transparent"
+        android:visibility="gone"
+        android:clickable="true"
+        app:layout_constraintTop_toTopOf="@id/act_collect_input_view"
+        app:layout_constraintBottom_toBottomOf="@id/toolbarBottom"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
+
 </androidx.constraintlayout.widget.ConstraintLayout>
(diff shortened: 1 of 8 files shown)


trife:
- [ ] Text traits are still editable
- [ ] Dismiss toasts before displaying new ones

chaneylc:
I fixed the text trait locking issue, tested with scanners too.

app/src/main/res/layout/activity_collect.xml
```diff
@@ -143,4 +143,16 @@
...
+        android:id="@+id/lockOverlay"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:background="@android:color/transparent"
+        android:visibility="gone"
+        android:clickable="true"
+        app:layout_constraintTop_toTopOf="@id/act_collect_input_view"
+        app:layout_constraintBottom_toBottomOf="@id/toolbarBottom"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toEndOf="parent" />
```
chaneylc, line 156: I originally went with implementing locking like this. I discussed this with @trife when I realized it disables scrolling and viewing captured images. Now it also disables repeated measures, and the metadata dialog. @trife let me know if this is what you wanted but originally we decided against this approach for these reasons. This approach really does not pair well with the existing repeated measures, unless we maybe constrain it below the collect input view when repeated measures is enabled.

trife: Let's try this approach and see if we get push back from users. The other approach seems more prone to error.

PR #990, opened 2024-06-26
Skip entries update and bugfix

### LINKED ISSUE #987: Skipping plots with missing data if the note was not the one most previously taken[REQUEST]:
### Description

When taking notes for two traits we would like to be able to jump to the next plot that is missing either trait not just skipping to the next plot that is missing data for the trait you currently have selected. For example when taking flowering time data we have silking date and anthesis date as our active traits but we also have stand counts loaded into Field Book as an unselected trait. We would like to skip to the next plot missing either silking OR anthesis data but currently if we select "Skip entries across active traits" if the last note we took was anthesis date it will skip to the next plot missing anthesis data even if there were plots in between that were missing silking date. Same if the last note was silking date it will jump to the next plot missing silking not catching any plots missing anthesis. If we select "Skip entries across all traits" it will skip no plots since stand count is in as a trait but had no saved notes currently. 

I don't think that we were having this same issue last year and wonder if it has just been since the issue "Return to first trait when advancing to next plot #722" was resolved

### Alternatives

_No response_

### Additional Context

_No response_

bellerbrock:
## Description

Fixes bug where skip entries for all traits preference would crash collect activity by attempting to set traitboxview to a trait that wasn't visible.

closes #987

## Type of change

## Checklist:

## Changelog entry

```release-note
Skip entries update and bugfix
```

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
@@ -1374,7 +1374,7 @@ public class CollectActivity extends ThemedActivity
                 startActivityForResult(intent, REQUEST_FILE_EXPLORER_CODE);
             }
         } else if (itemId == nextEmptyPlotId) {
-            rangeBox.setPaging(rangeBox.movePaging(rangeBox.getPaging(), 1, false, true));
+            rangeBox.setPaging(rangeBox.movePaging(rangeBox.getPaging(), 1, true));
             refreshMain();
         } else if (itemId == jumpToPlotId) {
             String moveToUniqueIdValue = preferences.getString(GeneralKeys.MOVE_TO_UNIQUE_ID, "");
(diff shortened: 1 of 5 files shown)


chaneylc (approved):
Looks good, maybe we should also think about updating the descriptions for that type of navigation. The user described the settings as "Skip entries across active traits" when really its "Skip entries across active trait" (trait not traits, where the third option is supposed to be traits) Their description was actually accurate to what is supposed to happen for that mode.

bellerbrock:
@chaneylc thanks. And agreed re the setting description. They have been updated in the v6 settings PR to `    <string name="preferences_behavior_skip_entries_title">Skip Entries</string>
    <string name="preferences_behavior_skip_entries_disabled">Disabled</string>
    <string name="preferences_behavior_skip_entries_with_data">Arrows skip entries that already have values for the current trait</string>
    <string name="preferences_behavior_skip_entries_across_all_traits">Arrows skip entries that already have values for all traits</string>`

PR #1009, opened 2024-07-18
v5 repeated values compatibility with brapi field sync

### LINKED ISSUE #548: [BUG] Test if multi measures work with BrAPI Sync

chaneylc:
fix #548
added repeated values to brapi sync feature

## Description

## Type of change

## Checklist:

## Changelog entry

```release-note

```

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java
+++ b/app/src/main/java/com/fieldbook/tracker/brapi/service/BrAPIServiceV2.java
@@ -32,6 +32,7 @@ import com.fieldbook.tracker.utilities.FailureFunction;
 import com.fieldbook.tracker.utilities.SuccessFunction;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
 import org.brapi.client.v2.ApiResponse;
 import org.brapi.client.v2.BrAPIClient;
@@ -106,6 +107,8 @@ import java.util.function.BiFunction;
 public class BrAPIServiceV2 extends AbstractBrAPIService implements BrAPIService {
 
     private static final String ADDITIONAL_INFO_OBSERVATION_LEVEL_NAMES = "observationLevelNames";
+
+    private final String REPEATED_VALUE_INDEX_KEY = "rep";
 
     //used to identify field book db id in external references
     private final String fieldBookReferenceSource = "Field Book Upload";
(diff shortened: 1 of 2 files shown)


bellerbrock:
@chaneylc looking at the [getExportDbData](https://github.com/PhenoApps/Field-Book/blob/f2fb39bca264181667f7dc93e1098f3a54f39590/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt#L157) function the export uses, should be straightforward to calculate rep in the SQL. Just adding `RANK() OVER (PARTITION BY obs.observation_unit_id, obs.observation_variable_name ORDER BY obs.observation_time_stamp) AS calculated_rep` to the select clause should do the trick.

chaneylc:
> @chaneylc looking at the [getExportDbData](https://github.com/PhenoApps/Field-Book/blob/f2fb39bca264181667f7dc93e1098f3a54f39590/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt#L157) function the export uses, should be straightforward to calculate rep in the SQL. Just adding `RANK() OVER (PARTITION BY obs.observation_unit_id, obs.observation_variable_name ORDER BY obs.observation_time_stamp) AS calculated_rep` to the select clause should do the trick.

That is definitely a good use of window functions but we can't do it this way. API 30 is when sqlite 3.28 was introduced which was the first version with window functions, we still have a minimum of 21. I'm guessing even android studio's sqlite version is below this for API 26 compatibility.

bellerbrock:
@chaneylc Gotcha, that would explain some of the syntax errors I kept getting when trying the window function approach in the observationDao (though for some reason it works fine in the observationUnitPropertyDao). 

Its not as concise but a subquery should also do the trick. But rather than repaeat it everywhere the observations table is accesses, what do you think about making an ObservationsDetail View? I see [a TODO here suggestiing one here](https://github.com/PhenoApps/Field-Book/blob/f2fb39bca264181667f7dc93e1098f3a54f39590/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt#L162C12-L162C50), I could add it and have rep calculated using a subquery.

chaneylc:
> @chaneylc Gotcha, that would explain some of the syntax errors I kept getting when trying the window function approach in the observationDao (though for some reason it works fine in the observationUnitPropertyDao).
> 
> Its not as concise, but a subquery should also do the trick. But rather than repeat it everywhere the observations table is accessed, what do you think about making an ObservationsDetail View? I see [a TODO here suggestiing one here](https://github.com/PhenoApps/Field-Book/blob/f2fb39bca264181667f7dc93e1098f3a54f39590/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationDao.kt#L162C12-L162C50), I could add it and have rep calculated using a subquery.

Yeah that sounds ok, it would be nice to combine that query and getHostImageObservations too. These were originally meant to return brapi observations, so be careful there with the data source null check to make this robust enough for local repeated measures.

bellerbrock:
@chaneylc  Ok, working on the view approach now. Looks like the existing views are defined in [Migrator.kt](https://github.com/PhenoApps/Field-Book/blob/main/app/src/main/java/com/fieldbook/tracker/database/Migrator.kt) then its imported into the DataHelper and Daos for view creation and view name retrieval, respectively. But the Migrator seems written to only be used for v8 to v9 db upgrade. Do you think I should refactor it so it can be used for handling additional views, or would you suggest a different approach?

chaneylc:
> @chaneylc Ok, working on the view approach now. Looks like the existing views are defined in [Migrator.kt](https://github.com/PhenoApps/Field-Book/blob/main/app/src/main/java/com/fieldbook/tracker/database/Migrator.kt) then its imported into the DataHelper and Daos for view creation and view name retrieval, respectively. But the Migrator seems written to only be used for v8 to v9 db upgrade. Do you think I should refactor it so it can be used for handling additional views, or would you suggest a different approach?

Instead of refactoring I think it would be best to encapsulate your view definition, and any functions necessary to maintain it in its own DAO class.

bellerbrock:
@chaneylc There were a lot of messy downstream effects when using the view, so I've thought better of it, reverted those commits, and have gone with calculating rep for the brapi obs based on the existing obs during the sync save process. Let me know what you think of this solution.

zrm22 (approved):

PR #1036, opened 2024-08-27
Field specific traits

### LINKED ISSUE #1014: [REQUEST]: Maintain Obs Vars list associated with DeltaBreed experimental dataset
### Description

**Current Behavior**

- No BrAPI endpoint for dataset: 
- DeltaBreed users create dataset(s) before data collection. Information from these datasets is pulled via BrAPI to Field Book to support data collection. However because there is no dataset concept, when multiple datasets are pulled the observation variable lists associated with each dataset is not respected. 
- Field Book exports dataset files after data collection (with obs) in two formats: tabular & database

**Limits of current behavior**

Pull two datasets from DeltaBreed into FB. 

- DS1 = 3 observations variables = A, B, C
- DS2 = 8 observations variables = A, B, C, D, E, F, G, H

When DS2  is loaded, all 8 observation variables are now erroneously selected for DS1.

### Alternatives

Desired Dataset Behavior

- When DS1 is loaded to Field Book the correct 3 observation variables are loaded and selected.
- When DS2 is loaded to Field Book the correct 8 observation variables are loaded and selected.

### Additional Context

Experimental (Trial) Dataset is not in BrAPI

- Observation units  and related details (independent experimental variables, aka experimental design) 
- Observation Variable list (dependent experimental variables)
-  Observations (optional, only available after data collection)

Pete doesn't think it needs to be in BrAPI

> Hey Shawn, Field Book has the ability to manually enable/disable traits. I think it would be relatively trivial for them to track ObsVars downloaded with a dataset, then enable/disable them automatically when a user changes fields in the app. This functionality would need to be added to Field Book to solve your issue either way, whether it uses the current BrAPI import methods, or if BrAPI had an explicit dataset endpoint that linked everything.
I would recommend opening an issue/feature request on the Field Book GitHub (if you haven't already) and perhaps copying this explanation into the [#field-book](https://plantbreedingapi.slack.com/archives/C01T9UK7EJC) channel to make sure the right people see it.I'm not sure I see the need for a new BrAPI endpoint right now, based on this use case. ObsUnits, ObsVars, and Observations can all be linked together and queried together by a trialDbId or studyDbId depending on how you organize your experiments. Field Book is currently retrieving all its data by studyDbId but it isn't maintaining the relationship between ObsVar and Study after pulling the data.

Duplicates BI-2249

bellerbrock:
## Description

Adds trait editor activity access from field detail page, with any visibility changes tracked on the field-level using a study_id jsonb column in the observation_variables table.

closes #1014

## Type of change

## Checklist:

## Changelog entry

```release-note

```


bellerbrock:
Not a finished product yet, opening as a draft PR as a place for testing/feedback

The main remaining question is if trait visibility should be field-specific everywhere now, or just when accessing the trait and collect activities from the field detail fragment? So far the changes are implemented with the second strategy in mind - preservsing the field-independent visibility info when accessing the trait and collect activities from the config activity. But I'm worried that both approaches side-by-side will be confusing, and it would be a lot simpler to completely switch over.

If we don't completely switch over I will still need to go through and update additional places that check trait visibility (summary, datagrid, export) to determine in which contexts they should apply the field-specific trait visibility or the general trait visibility.

I also still need to set field-specific visibility for traits imported by Brapi sync.

chaneylc (changes requested):
Overall this looks good, but I'm confused why we are disabling trait editor with no field selected if we're just copying visibilities, and I think it would be interesting to copy trait definitions instead. That way each specific page would have a smaller list of traits, and maybe they could be individually deleted/managed.


1. I was a little confused with the trait duplication button. Minor thing: I think using a list dialog instead of a spinner might be better here. 
2. Something odd with the spinner duplicate choice. I think having the icon there is cool but I'm not sure how much it adds and I thought it was confusing. This shows the number of visible traits, but the icon and use of that is unclear (I think users will be confused to see 0's when traits are actually defined but set to invisible). I think it would be interesting if instead of visibility this duplicates trait definitions, then instead of a trait count you see a +/- difference of traits or at least a difference in visibility/sorted order.
3. if the idea is to copy trait visibility / order, I think it's still a valid usecase to use trait editor without a loaded field to manage the master list. I think it's odd that you can create a trait in a 'field specific' page and that will be created in all field specific pages, similarly I thought deleting traits in a field-specific page would keep the trait for other fields.

trife:
@chaneylc What you're describing is the long term vision but the approach in this PR is a stopgap since this has been a recurring issue for several programs. The second iteration of this will require modifications to make TraitEditorActivity a master/fixed list instead of adjustable and will require a new approach to add traits from the master list to the field-specific lists. That being said, we should definitely add a dialog that tells the user that deleting traits will delete them from every field and/or remove the option to delete from the field-specific list.

Issue #574, opened 2023-05-05
Primary/Secondary Order modifications


trife:
- [x] Replace textviews/data in Collect with InfoBars
- [x] Check any queries that use these to ensure they can dynamically retrieve the value from the user-selected Primary/Secondary
- [x] Attempt to automatically select these using previous selections. If no previous selections, try row/column and then range/plot to automatically set
- [x] OnPress - open alertdialog that contains quickGoTo functionality
- [x] Remove quickgoto setting
- [x] Remove primary/secondary from the import process and all queries

PR #1048, opened 2024-09-13
BrAPI Workflow UI

chaneylc:
## Description

New UI for BrAPI study import

- [x] Documentation
- [x] Minor refactoring: package, code duplication in BrAPIServiceV2
- [x] Various string resources
- [ ] ~~Improve sorting with toolbar / maybe better table view with column choices, currently sorts by list item label name ascending order~~
- [x] Discuss / improve search text filtering, when choosing/filtering studies users can add text segments, these text segments are logically matched using "OR" to filter models. Searching is different for non-study objects such as programs, these search texts are not saved. **Should be AND**
- [x] Discuss / improve filtering, non study filters such as programs do not affect other filters, I could see it being useful to choose a program then when browsing trials have only those belonging to the program show up **Filters should affect other filters**
- [x] Add time-based cache invalidation preference
- [x] Discuss cache persistence: currently cache is reset by the user in the toolbar UI, or whenever the server is authenticated, another option is to keep a directory hierarchy of caches for each server visited
- [x] Add exception UI
- [x] Discuss / improve level/sort/prim/sec UI, currently I have it choose "plot", "Row", "Col" if they exist and leave sort off by default.
- [x] level / unit handling. Do many studies have data across different levels? There could be optimizations for getting this data if so.
 
Required BrAPI calls:
GET studies / trials are required, latter is needed to filter by program since program data is not included in study objects
GET commonCropNames, trials, seasons, programs
GET observationlevels (although I noticed servers implement this differently, some don't return anything when supplying a programDbId, in that case you have to supply null)
GET observationVariables (from previous implementation, traits are imported with the studies)
GET observationUnits
GET germplasms (used to populate unit attribute table further)

Tested on BrAPI test server and Cassavabase.
The test server seems to have only one study that contains plots.
Cassavabase has a lot of data, making filtering interesting, but most calls take a long time to load.

## Type of change

- [x] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)

```release-note

New BrAPI Study import UI
```


trife:
- [ ] Rename fragment name when importing study. Currently says "Field Book"
- [ ] Convert data elements to chips in the imported study list item (currently plots, traits, location)
- [ ] For level selection- is there a way to see if these levels exist for a selected study and remove those that don't?
- [ ] Use consistent button/UI for "Import" action (study chooser fragment vs. study chosen fragment)
- [ ] Move active filters to below "Search" widget and out of pop up card
- [ ] With above change, I'm not sure if popup card is needed. Maybe just a dialog with a list of filter options instead?
- [ ] Search widget has some weird corner shading
- [ ] Include trial name in study list items
- [ ] Include the results number in the scrollview so it disappears when scrolling through the list. Maybe just include this as a chip so it matches the rest of the UI when filters are present?
- [ ] The actual import process is really slow, even for studies with a small number of entries (testing on Cassavabase). Is there anything we can do on our side or is due to issues on their side?
- [ ] Could also move the semi-opaque import process to a list of checkmarks when each step is done
- [ ] Primary/secondary selection can eventually be removed. It might be worth working on #574 before this gets merged

trife (changes requested):
- [ ] Change font color on Filter/Import/Save button to black (currently white)
- [ ] I set cache invalidation to daily but the activity attempted to download all of the studies when I tried importing a second field.
- [x] Add confirmation prompt to cache invalidation
- [ ] Is there any way to make the indeterminate loading bar determinate?
- [x] Move invalidate cache toolbar icon to overflow menu
- [ ] Move search result count to search bar hint (e.g. `Search... 524 results`)
- [ ] Include count in filter pages (in search bars)
- [ ] Move clear filters icon to toolbar. Only show if there are filters applied
- [ ] Move Sort tab to last position
- [ ] All pages say "Loading trials" but should say "Loading studies/seasons/trials/programs"
- [ ] Should we add any other pieces of info to the returned study list elements?
- [ ] On the Import page, I'm getting a perpetual loading wheel in the study list item and no traits are being loaded. Not sure if this is an issue on our end or cassavabase's.

@lukasmueller - loading studies from cassavabase via brapi is very slow. I'm not sure if this is something we can optimize on our end or if it's something that should be worked on by the cassavabase team.

Issue #1055, opened 2024-10-02
[BUG]: BrAPI field import germplasm attributes not imported in some cases


nickpalladino:
### Description

As part of the BrAPI field import process a germplasm search call is made and the results are used to populate the pedigree and synonyms attributes for display on the collect page. When the search call responds with a searchResultDbId and a GET call is made to get the results by Field Book, if the results are not ready by the time the first call is made they will not be imported. This seems to be because the retry logic in Field Book is not handling that case properly and is only making the single request. Problem seems to be in the `BrAPIServiceV2::executeBrapiSearch` while loop exiting before it should.

chaneylc:
@nickpalladino on our latest version we are testing out a new brapi import, can you test to see if the problem still occurs? You can enable it in the Settings/Experimental/New BrAPI Import UI. Currently that implementation uses the germplasmGetAsync call with the chosen study db id, if that method doesn't work for you I can implement the search case. Thanks.

trife:
@nickpalladino Have you been able to test if this is still an issue?

nickpalladino:
@trife I have not, I'll see if I can get things set up to test this soon

trife:
I'm closing this but feel free to reopen if the issue pops up again.

PR #1093, opened 2024-12-04
brapi_germ_accession_number

chaneylc:
### Change Type

### Release Note

```release-note
Accession number now included for germplasm imported via BrAPI
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**
- [x] **`WAIT`** (No immediate release, but the changelog will be still be updated if there is a release note.)

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/brapi/io/BrapiStudyImportActivity.kt
+++ b/app/src/main/java/com/fieldbook/tracker/activities/brapi/io/BrapiStudyImportActivity.kt
@@ -417,6 +417,10 @@ class BrapiStudyImportActivity : ThemedActivity(), CoroutineScope by MainScope()
 
                 germs.firstOrNull { it.germplasmDbId == unit.germplasmDbId }?.let { germ ->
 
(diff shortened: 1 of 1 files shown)


nickpalladino:
@chaneylc Is this part of the new experimental brapi import that was mentioned in https://github.com/PhenoApps/Field-Book/issues/1055? If so, is the plan to move everything to that? Does it need an equivalent change in the old brapi service code?

chaneylc:
These changes are only for the new brapi import. We can make the changes to the old brapi code if necessary but the idea is to replace it.

trife:
@nickpalladino We're going to try to avoid making too many more updates to the legacy brapi import. The (currently) experimental version should graduate to a release feature before our next minor release (likely before EOY). We're just working on adding a few final touches and allowing some partners to test it out.

nickpalladino:
Thanks for the update. We'll do some testing with the new brapi import code. I'm good with the change just going in the new code since the plan is to switch over to that soon.

trife (approved):

PR #1095, opened 2024-12-04
brapi_shared_study_names

chaneylc:
### Change Type

### Release Note

```release-note
Fields with different BrAPI study IDs can now have the same name
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife (changes requested):
I don't think we should include the study dbid in the listitem.

chaneylc:
I can remove it, but then we need a way to differentiate between studies with the same name.

trife:
Duplicate or non-descriptive names is against the recommended pattern for Field Book but isn't against the brapi spec. We can eventually group brapi studies by trial to side step this, but it's not a priority for now.

trife (approved):

PR #1125, opened 2024-12-17
field_search

chaneylc:
### Change Type

### Release Note

```release-note
Search box added to field list
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife:
The only addition that would make this better would be an added X symbol (like in settings search) that clears out whatever has been typed in. It might be worthwhile to add that to all of the search boxes, though.

Otherwise it seems good to me.

trife:
One thing that might be worth doing is to only display the search box if the list extends beyond the length of the display.

trife (approved):

PR #1131, opened 2025-01-07
remove_primary_secondary_choice

chaneylc:
### Change Type

### Release Note

```release-note

```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife (changes requested):
- [ ] Rest of the issues here: https://github.com/PhenoApps/Field-Book/issues/574
- [x] When changing selected primary/secondary, the replacement isn't formatted the same (truncated with colon) until activity restarts
- [x] When primary or secondary is ellipsized, there's a small added left margin
- [x] Truncate prefixes to the first three letters and shift the data over so more can be viewed on smaller screens
- [ ] The prefix chooser dialog should have a title (similar to the infobars)

trife:
@chaneylc I tested this branch with BrAPI and it worked with the test server and cassavabase but not the [deltabreed sandbox](https://sandbox.breedinginsight.net/).

chaneylc:
What didn't work for the deltabreed sandbox? Just tested it and I was able to import both of their studies, looks like the traits and units were imported correctly as well.

trife:
- [ ] Rename quick goto dialog title to `Quick GoTo`
- [ ] Remove icon in dialog
- [ ] Don't auto fill dialog with values
- [ ] Auto select edittext onFocus
- [ ] Search logic should only attempt to match both primary/secondary values if both are populated (this is the current behavior)
- [ ] Revert gravity changes to what's on main

trife (changes requested):
- [x] Rename quick goto dialog title to `Quick GoTo`
- [x] Remove icon in dialog
- [x] Don't auto fill dialog with values
- [x] Auto select edittext onFocus
- [x] Search logic should only attempt to match both primary/secondary values if both are populated (this is the current behavior)
- [ ] Revert gravity changes to what's on main

PR #1135, opened 2025-01-13
Bugs v6.0.3

chaneylc:
### Description

Various bug fixes

### Change Type

### Release Note

```release-note
Numerous bug fixes and enhancements
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


chaneylc:
> Looks good to me, I tested the affected features to the extent I can (settings, field delete, stat page, print trait, etc) and everything seems to work as expected.
> 
> Not sure if this was meant to be fixed in this PR, but I did notice that I can create a photo trait with special characters in the name, but then the photos fail to save or be exported (no error message though). Also this may be specific to my old device, but the new photo trait doesn't work by default on my pixel 4a, I have to switch it to android camera app.

Good catch, I found more usages of what's causing a recurring bug in the export code and the data grid. Exports should work with special character traits now, data grid should also work correctly now too.

PR #1138, opened 2025-01-18
Trait modifications

### LINKED ISSUE #270: Add indicator that data has been saved in addition to color

### LINKED ISSUE #414: Update trait layouts to consistently display the data input box for all traits
I think this should be broken into  three sub categories: traits that collect data, traits that collect media, and traits that perform actions. Traits that collect data should always show the current value in the cvET. Traits that collect media should display a standardized media view. Traits that perform actions can do whatever but may overlap with data/media depending on the action.

- [x] Move date trait layout out of current value
- [x] Audio trait update to media view

### LINKED ISSUE #519: Remove date save button and auto save on +1/-1

### LINKED ISSUE #1028: Date trait save button will overwrite data with current date even if data is already present
Related to #519

kamathprasad9:
### Description
Closes #270
Closes #1028 
Closes #414 

Italic for new/edited observation, red and bold for previously saved observations
Remove Date trait text out of the current value ET.

### Change Type

### Release Note
```release-note
Added indicators for edited vs saved data
```

___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife:
For the date trait:
- [ ] Move textview down to between +/- buttons
- [ ] Use one of the styles currently used in other trait views for the textview size/styling
- [ ] Move save button to second row- creating a three on top, two on bottom layout

chaneylc:
Deleting an NA value from the audio trait leaves an empty grey box until the layout is refreshed.

chaneylc:
It looks like repeated measures follows the same text flow, but on closer inspection the date trait doesn't italicize the text while other traits do.

chaneylc:
Percent trait layout has a bold text when seeking but italicized when finished seeking. I think it should remain italicized if possible.

kamathprasad9:
@chaneylc for the repeated measures, I was thinking to have three different sets of indicators
*  italic - edited for current measure
*  Bold and black - edited previously (for the trait) without navigating out of the trait/plot 
*  Bold and red - previously saved

Basically, italic is only the new indicator to show that current measure was edited; the rest two are the same as before. Let me know what you think

chaneylc:
I don't think this solves https://github.com/PhenoApps/Field-Book/issues/519 yet.

From my understanding, we want a systematic rework of when observation data is saved. From the issue it specifically says when some form of navigation occurs. Currently, value saving differs between traits.

On reviewing the PR it looks like data saving has remained the same. To tackle this, I think we can expand on the base trait layout abstraction to include a "save data" function that will be called from the collect controller whenever navigation occurs, and probably if the app closes as well.

chaneylc:
> @chaneylc for the repeated measures, I was thinking to have three different sets of indicators
> 
>     * italic - edited for current measure
> 
>     * Bold and black - edited previously (for the trait) without navigating out of the trait/plot
> 
>     * Bold and red - previously saved
> 
> 
> Basically, italic is only the new indicator to show that current measure was edited; the rest two are the same as before. Let me know what you think

This makes sense. I first thought italicized meant that it was an edited value that has not been saved yet, which might be true after the 519 changes.

kamathprasad9:
> I don't think this solves #519 yet.
> 
> From my understanding, we want a systematic rework of when observation data is saved. From the issue it specifically says when some form of navigation occurs. Currently, value saving differs between traits.
> 
> On reviewing the PR it looks like data saving has remained the same. To tackle this, I think we can expand on the base trait layout abstraction to include a "save data" function that will be called from the collect controller whenever navigation occurs, and probably if the app closes as well.

Although #519 specifies it differently, we had decided to treat this as a change only to the visual aspects rather than reworking the update observation workflow. 

@trife could you please confirm about the approach towards this issue?

trife:
Merged into https://github.com/PhenoApps/Field-Book/pull/1169/

PR #1153, opened 2025-01-31
crop_image_feature

chaneylc:
### Description

Photo traits can now define a crop region.

Workflow:
User must enable crop in trait definition, right now it is set to false as default. Like other trait parameters, this cannot be edited if data for that trait is already captured.

When observing data for that photo trait, the user can click the settings and define a crop region, or a dialog will show if the trait requires cropping and no region is defined when a photo is taken. In the latter case, the photo will be cropped and saved.

This feature introduced CropImageView which I added a fragment controller implementation along with an activity wrapper for the fragment. This is a modular and independent view, but requires two intent-based parameters, traitId (internal db id of the trait to update the crop region) and imageUri for the image to preview as it is cropped.

UI:
The CropImageView has a couple functionalities:
1. Users can drag the 4 boundary handles (top, bottom, start, end) within the bounds of the image to adjust ROI size
2. Users can drag the middle handle to move the entire region
3. Users can click save to save the rectangle coordinates to preferences for that trait
4. Users can click copy to save the rectangle coordinates to clipboard
5. Users can click reset to reset coordinates to the currently saved preferences
6. Users can click expand to reset coordinate to the image dimensions
7. Users can click/edit coordinate edit text, but coordinates will not be updated/invalidated unless the ACTION_DONE ime is pressed (the enter button)

Working with multiple devices:
User 1: set ROI, copy coordinate text and send to User 2
User 2: paste roi coordinates in edit text, press enter and save

- [ ] Needs testing on non-system cameras.
- [ ] Needs to disable crop setting button for traits that have it disabled.

https://github.com/user-attachments/assets/e8f68db7-df09-4071-b1d3-40b2f24c47dc

### Change Type

### Release Note

```release-note

```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -227,7 +227,10 @@ dependencies {
     implementation "androidx.exifinterface:exifinterface:1.3.6"
 
     //camerax
-    def camerax_version = "1.3.3"
+    def camerax_version = "1.4.1"
+
+    implementation "androidx.camera:camera-effects:$camerax_version"
+
     implementation "androidx.camera:camera-camera2:$camerax_version"
     implementation "androidx.camera:camera-lifecycle:$camerax_version"
     implementation "androidx.camera:camera-view:$camerax_version"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -206,6 +206,10 @@
                     android:scheme="fieldbook" />
             </intent-filter>
         </activity>
+        <activity android:name=".activities.CropImageActivity"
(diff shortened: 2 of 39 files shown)


trife (changes requested):
- [x] Captured images are not cropped but include the crop mask
- [ ] Cropping seems to happen from the internal border of the cropping area so there is a sliver of cropped area on the top and bottom (although one image I took didn't have this)
- [ ] Image delete button not aligned with thumbnails
- [ ] Trait-specific crop setting seems to be ignored (i.e. I can crop images from any photo traits)

PR #1165, opened 2025-02-18
Editable barcode search

### LINKED ISSUE #792: Add per field option to select which data column should be used for barcode search

bellerbrock:
### Description

Adds search attribute pref for changing column searched by move to unique id. Adds chip for editing pref on field detail page, limits options to unique attributes

closes #792

### Change Type

### Release Note

```release-note
Barcode attribute can now be selected on a per field basis
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


chaneylc:
@bellerbrock can you merge this with main before I review? Currently it's un-testable due to the ffmpeg libs breaking the gradle build. Merging with main should fix all of those issues.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -1992,57 +2128,42 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-                            int studyId = studyObj.getExp_id();
-                            String fieldName = studyObj.getExp_alias();
-
-                            String msg = getString(R.string.act_collect_barcode_search_exists_in_other_field, fieldName);
+                        IntentResult plotSearchResult = IntentIntegrator.parseActivityResult(resultCode, data);
+                        barcodeValue = plotSearchResult.getContents();
+                    }
 
-                            SnackbarUtils.showNavigateSnack(getLayoutInflater(), findViewById(R.id.traitHolder), msg, R.id.toolbarBottom,8000, null,
-                                (v) -> switchField(studyId, null));
```
chaneylc, line 2036: Correct me if I'm wrong, but this barcode search used to find the entry even if it was in another field. It looks like your changes remove that.

app/src/main/java/com/fieldbook/tracker/activities/FieldDetailFragment.kt
```diff
@@ -186,6 +191,15 @@ class FieldDetailFragment : Fragment(), FieldSyncController {
             }
         }
 
+        editUniqueChip.setOnClickListener {
+            fieldId?.let { id ->
+                val field = database.getFieldObject(id)
```
chaneylc, line 196: This is querying on the main thread, try to avoid it and use coroutines.

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
```diff
@@ -941,12 +941,148 @@ public boolean moveToSearch(
             }
         }
 
+        if (command.equals("barcode")) {
+            int currentFieldId = preferences.getInt(GeneralKeys.SELECTED_FIELD_ID, 0);
+            Log.d("Field Book", "Barcode search in current field: " + currentFieldId + ", searching for: " + data);
+
+            ObservationUnitModel[] matchingUnits = database.getObservationUnitsBySearchAttribute(
+                    currentFieldId, data);
+            Log.d("Field Book", "Search attribute results: " + matchingUnits.length + " units found");
```
chaneylc, line 950: Would it be too much work to rewrite this portion to use the BackgroundUITask? Otherwise this is a query on the main thread.

chaneylc (changes requested):

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
bellerbrock: Not removed, just moved into its own function `searchAcrossAllFields` and called at the end of the moveToSearch logic if no match is found in the current field. I also expanded it so it checks against both the searchAttributes and the uniqueIds of other fields.

trife (approved):

app/src/main/java/com/fieldbook/tracker/activities/CollectActivity.java
bellerbrock: @chaneylc I spent some time working on this, at first ran into the problem of the utility not being able pass the search results from the background thread to the main thread. Based on our discussion on slack I also tried adding an a 'Any' return type to backgroundBlock and an 'Any' parameter to the uiBlock, but that then broke the other places the utility is used in dialogs.

Definitely a good idea to move the queries of the main thread, but I'm thinking it'd be best to tackle after the merge as part of larger refactoring. And/or with your assistance to expand the utility in a less disruptive way.

PR #1186, opened 2025-03-21
Device name preference

### LINKED ISSUE #877: Add customizable device name setting to profile settings
This doesn't need to get saved to the observation table but should be joined to database/long format exports and displayed in Nearby share. Can default to the device model.

kamathprasad9:
### Description
Create a new device name preference in Profile
Closes #877 

### Change Type

### Release Note

```release-note
Device name can now be customized in Profile Settings
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife (changes requested):
- [x] Make a separate settings section in Profile called `Device`
- [x] Use [this](https://pictogrammers.com/library/mdi/icon/cellphone-information/) icon for the pref
- [x] If the default has been adjusted, can we add the device model number to places where the name is being used? e.g. `TR's Phone (Pixel 6)`

chaneylc (dismissed):
Looks good. Didn't catch this before but this line

Utils.makeToast(context, String.format(getString(R.string.nearby_share_file_imported_successfully), fileName))

formatted the string incorrectly for me and I saw something like
-871235921501259234 transferred successfully

We should probably limit the device naming like Bryan did the field renaming feature. Since this is in the csv output now, it will cause issues if they contain special characters.

trife (approved):

chaneylc (approved):

PR #1191, opened 2025-03-25
Export only active trait media files

### LINKED ISSUE #995: [BUG]: Bundling media does not respect the "active traits only" option in the export dialog
### Description

[BUG]: when export bundle carry all the pictures ever done for that import field, huge zip files

### Expected behavior

Only export the photo folder Im saying to export not all the pictures attach to that field 

### Field Book version

Latest

kamathprasad9:
### Description
Fixes #995 

### Change Type

### Release Note

```release-note
Bundled media is now exported as expected for active and existing traits
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/utilities/ExportUtil.kt
+++ b/app/src/main/java/com/fieldbook/tracker/utilities/ExportUtil.kt
@@ -73,6 +73,10 @@ class ExportUtil @Inject constructor(@ActivityContext private val context: Conte
     private val timeStamp = SimpleDateFormat("yyyy-MM-dd-hh-mm-ss", Locale.getDefault())
     private var multipleFields = false
 
+    // a temporary directory is created when activeTraits is selected
+    // this needs to be deleted AFTER zipping is completed
+    private var tempDirectory: DocumentFile? = null
+
     fun exportMultipleFields(fieldIds: List<Int>) {
         this.fieldIds = fieldIds
         this.multipleFields = fieldIds.size > 1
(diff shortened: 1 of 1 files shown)


trife (changes requested):
@kamathprasad9 When bundling images, the total list of traits should be limited to those that currently exist, so the bundle code should be modified to take a list of media traits to include and adjust that list based on user choice.

trife (approved):

app/src/main/java/com/fieldbook/tracker/utilities/ExportUtil.kt
```diff
@@ -483,11 +487,56 @@ class ExportUtil @Inject constructor(@ActivityContext private val context: Conte
...
+        tempDirectory = exportDir?.createDirectory(tempDirName)
+
+        try {
+            tempDirectory?.let { tmpDir ->
+                // create study name directory
+                val studyDir = tmpDir.createDirectory(studyName)
+
+                studyDir?.let { studyDirectory ->
+                    mediaDir.listFiles().forEach { traitDir ->
+                        val traitDirName = traitDir.name ?: ""
```
chaneylc, line 521: Use FileUtil.sanitizeFileName on the trait name in case it has special characters that the FS can't handle.

chaneylc (changes requested):
Requesting minor change, otherwise it looks good. I tested by activating/deactivating a picture trait and it was correctly exporting only when the trait was active.

chaneylc (approved):

PR #1192, opened 2025-03-26
Fix db export for only unique identifier

### LINKED ISSUE #1177: [BUG]: Local export of data from BrAPI-imported field is missing columns
### Description

Field trial was imported from BreedBase via BrAPI. Some datapoints were collected on trait also imported via BrAPI. Instead of exporting via BrAPI, local export was used for: Database format, Only unique identifier, and Only active traits. The exported file included the germplasm name as ObservationUnitDbId, EntryType as trait name, ObservationUnitDbId as value, plot_name as timestamp, plot_number as person, block as location, and Rep as number. Basically, mixed up and missing data. There didn't seem to be any problems when you export all imported columns. 

File showing incorrect export: IncorrectExport_database.csv
File showing correct column from exporting all import columns: AllImportedColumnsExport_database.csv

[IncorrectExport_database.csv](https://github.com/user-attachments/files/19188625/IncorrectExport_database.csv

[AllImportedColumnsExport_database.csv](https://github.com/user-attachments/files/19188660/AllImportedColumnsExport_database.csv)

### Expected behavior

Expected the exported file to only contain the correct values for "ObservationUnitDbId","trait","value","timestamp","person","location","number". 

### Field Book version

6.0.8

kamathprasad9:
### Description
Fixes #1177 

### Change Type

### Release Note

```release-note
Fixed a bug where column values were being assigned to the wrong header
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt
+++ b/app/src/main/java/com/fieldbook/tracker/database/dao/ObservationUnitPropertyDao.kt
@@ -226,6 +226,11 @@ class ObservationUnitPropertyDao {
                     addAll(traitRequiredFields)
                 }.toTypedArray()
 
+                // get the column indices for the requiredColumns from fullCursor
+                val columnIndices = requiredColumns.map { columnName ->
+                    fullCursor.getColumnIndex(columnName)
+                }
+
                 // Create and return a cursor that only includes the required columns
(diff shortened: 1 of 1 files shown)


trife (changes requested):
Looks like this added a new issue. In the database export, it's exporting data from all of the columns that should be excluded.

PR #1206, opened 2025-03-29
Show previous person names

### LINKED ISSUE #1045: [REQUEST]: Offer previously used names as options when changing the person name
### Description

Possible implementation could be to append each distinct person name entered in the app to a list in preferences. Or just retrieve the distinct values from the collector column in the observation table.

### Alternatives

_No response_

### Additional Context

_No response_

kamathprasad9:
### Description
Closes #1045 

### Change Type

### Release Note

```release-note
User names are now saved in a list for rapid switching
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/preferences/ProfilePreferencesFragment.java
+++ b/app/src/main/java/com/fieldbook/tracker/preferences/ProfilePreferencesFragment.java
@@ -11,6 +11,7 @@ import android.widget.EditText;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
+import androidx.core.content.ContextCompat;
 import androidx.preference.Preference;
 import androidx.preference.PreferenceFragmentCompat;
 
@@ -53,8 +54,6 @@ public class ProfilePreferencesFragment extends PreferenceFragmentCompat impleme
 
         nameManager = new PersonNameManager(preferences);
 
-        previouslySavedNames = nameManager.getPersonNames();
-
         ((PreferencesActivity) this.getActivity()).getSupportActionBar().setTitle(getString(R.string.settings_profile));
 
         profilePerson = findPreference("pref_profile_person");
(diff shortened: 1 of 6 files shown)


trife (changes requested):
This is a great start!

- [x] If >1 name stored, change personPref behavior to show list first
- [ ] Add 'New Person' button to previous names dialog
- [ ] Add confirmation dialog before clearing
- [x] Change reminder default behavior to open up list first
- [ ] Fix layout of reminder (bottom has too much padding)
- [ ] Retitle previous names dialog to "Previous Names"

trife (changes requested):
- [ ] Let's show the person list after the size == 1 instead of >1

app/src/main/java/com/fieldbook/tracker/preferences/ProfilePreferencesFragment.java
```diff
@@ -73,24 +92,31 @@ public void onCreatePreferences(Bundle savedInstanceState, String rootKey) {
             boolean updatePerson = arguments.getBoolean(GeneralKeys.PERSON_UPDATE, false);
 
             if (updatePerson) {
-
-                showPersonDialog();
-
+                previouslySavedNames = nameManager.getPersonNames();
```
chaneylc, line 95: Looks like duplicated code, this should be a function.

app/src/main/java/com/fieldbook/tracker/utilities/PersonNameManager.kt
```diff
@@ -0,0 +1,139 @@
...
+     */
+    fun getPersonNames(): List<PersonName> {
+        val namesList = mutableListOf<PersonName>()
+
+        try {
+            val namesArray = getCurrentNamesArray()
+
+            for (i in 0 until namesArray.length()) {
+                val nameObj = namesArray.getJSONObject(i)
+                val firstName = nameObj.getString("firstName")
```
chaneylc, line 76: It's worth defining these json keys as constants.

chaneylc (approved):
Approved with minor code improvement suggestions.

The previous name dialog should have the names align with "Clear", and a way to individually delete names will be useful.

I think a better UX would be utilizing preference components such as a list preference for the names, instead of using dialogs but that's just me :P

trife (approved):

PR #1215, opened 2025-04-04
Field grouping/archiving

### LINKED ISSUE #163: [REQUEST] Archive/Group fields by attributes

### LINKED ISSUE #177: [REQUEST] Allow study grouping by trial

kamathprasad9:
### Description
Closes #163 
Closes #177 

### Change Type

### Release Note

```release-note
Fields can now be grouped and archived
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


kamathprasad9:
- [ ] Show archived group as a list item 
- [ ] Fix searching behavior

trife (changes requested):
- [ ] Check menu styling- overflow is showing some odd styling
- [x] Add archive as a menu level option instead of being nested in groups dialog
- [ ] Add option to disable groups to top toolbar
- [ ] Sort by name should also sort groups by name
- [ ] Move archive to end of list (and it shouldn't get sorted)
- [ ] Long click on group to select all fields within

trife (changes requested):
- [ ] Use [this](https://pictogrammers.com/library/mdi/icon/group/) icon for the top toolbar
- [ ] Only display show/hide groups once a group has been created
- [x] Don't expand all groups when performing group actions (e.g. moving a field to a new group or archiving)
- [x] Related to above: is it possible to save expanded/collapsed state for groups?
- [x] Scroll to top when enabling groups
- [x] Expand the group when longpressing to select all
- [x] Include the number of group members in parentheses after the group name
- [ ] Add the group name (if it exists) as a chip in the field details page
- [x] Let's move ungrouped to bottom (above archived)
- [ ] Add [collapse all](https://pictogrammers.com/library/mdi/icon/arrow-collapse-vertical/) / [expand all](https://pictogrammers.com/library/mdi/icon/arrow-expand-vertical/) groups (always in overflow menu)

trife (changes requested):
- [ ] Selecting all when some fields are archived results in a crash (`java.lang.NullPointerException: Attempt to invoke virtual method 'int com.fieldbook.tracker.objects.FieldObject.getExp_id()' on a null object reference`)
- [ ] When a user creates a group, toggle group visibility
- [ ] If the group name and trial name are the same, hide the group name chip in the field details page
- [ ] Remove the icon from the archived list item

trife (changes requested):
- [x] Left margin of Archived should line up with other list items and height should match other list items
- [x] Select all should select all non-archived fields and expand groups. Currently only selects fields in groups that are expanded

trife (approved):

app/src/main/java/com/fieldbook/tracker/activities/FieldArchivedActivity.kt
```diff
@@ -0,0 +1,323 @@
...
+            val layoutParams = ActionBar.LayoutParams(
+                ActionBar.LayoutParams.WRAP_CONTENT,
+                ActionBar.LayoutParams.WRAP_CONTENT
+            )
+            customTitleView?.layoutParams = layoutParams
+
+            mode.customView = customTitleView
+
+            return true
+        }
```
chaneylc, line 141: Update the menu implementation like I did in FieldEditorActivity using onPrepareOptionsMenu

app/src/main/java/com/fieldbook/tracker/activities/FieldEditorActivity.java
```diff
@@ -519,9 +450,115 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
+                m.setAccessible(true);
+                m.invoke(menu, true);
+            }
+            catch(NoSuchMethodException e){
+                Log.e(TAG, "onMenuOpened", e);
+            }
+            catch(Exception e){
+                throw new RuntimeException(e);
+            }
+        }
```
chaneylc, line 468: Comment out this feature for now since it isn't consistent with other app menus.

app/src/main/java/com/fieldbook/tracker/activities/FieldEditorActivity.java
```diff
@@ -1113,4 +1209,129 @@ public void afterTextChanged(android.text.Editable s) {
...
+        boolean anyFieldsInGroup = false;
+
+        for (Integer fieldId : fieldIds) {
+            FieldObject field = fieldList.stream()
+                    .filter(f -> f.getExp_id() == fieldId)
+                    .findFirst()
+                    .orElse(null);
+
+            if (field != null) {
+                String groupName = StudyGroupDao.Companion.getStudyGroupNameById(field.getGroupId());
```
chaneylc, line 1227: Activity/Fragment code should only call queries with DataHelper

app/src/main/java/com/fieldbook/tracker/adapters/FieldAdapter.java
```diff
@@ -231,23 +360,290 @@ public void onBindViewHolder(ViewHolder holder, int position) {
...
+
+        // add group-field entries
+        for (Map.Entry<String, List<FieldObject>> entry : groupedFields.entrySet()) {
+            String groupName = entry.getKey();
+            List<FieldObject> groupFields = entry.getValue();
+
+            if (groupFields.isEmpty()) continue;
+
+            Integer groupId = StudyGroupDao.Companion.getStudyGroupIdByName(groupName);
+            boolean isExpanded = groupId == null || StudyGroupDao.Companion.getIsExpanded(groupId);
```
chaneylc, line 486: DAOs should not be used outside the DataHelper class, create an interface that defines the required functionality.

app/src/main/java/com/fieldbook/tracker/database/Migrator.kt
```diff
@@ -641,4 +667,12 @@ class Migrator {
...
         }
     }
+
+    class StudyGroup private constructor() {
+        companion object Schema {
+            const val PK = "id"
+            const val FK = "group_id"
+            const val TABLE_NAME = "study_groups"
+        }
+    }
```
chaneylc, line 677: Expand on this. You can use my nix trait sensor as an example. Define all column names and ensure they are not camel case. Update queries with the newly defined static variables.

app/src/main/java/com/fieldbook/tracker/database/dao/StudyGroupDao.kt
```diff
@@ -0,0 +1,143 @@
...
+                StudyGroup.TABLE_NAME,
+                select = arrayOf("isExpanded"),
+                where = "${StudyGroup.PK} = ?",
+                whereArgs = arrayOf("$groupId")
+            ).toFirst()["isExpanded"] as? String
+
+            result != "false"
+        } == true
+    }
+}
```
chaneylc, line 143: There are many usages of getColumnIndexOrThrow. This could potentially cause an exception and app crash. Work on exception flow/capturing and notifying the user of an error rather than causing a crash.

chaneylc (changes requested):
Requesting some minor code changes, otherwise it looks good. Great job!

chaneylc:
I think there is still a bug where if you make a selection of fields you can no longer go to the field detail page. There needs to be some logic to cancel the selection mode in both the fields and archive page.

chaneylc:
@trife should you be able to archive the active field? Right now you can archive it and it remains the active field. 

@kamathprasad9 these changes should also be made in the archive fragment, same bug occurs there where you can't go to the field detail anymore after making a selection.

kamathprasad9:
I had the same question in mind about archiving active field - I was thinking we should allow archiving of active field, but upon archiving we will end up having no active field (similar to how delete field works). What do you think @chaneylc

chaneylc:
After reviewing some of the linked issues, I think when a field is active and archived we should warn the user with a dialog to confirm archiving. If it is archived the field should be deselected and there should be no active field until another is explicitly chosen.

chaneylc:
When a field is in a group, the list item icon does not update when clicked (but the field is set to active) until you refresh the screen (going to config and back to field editor).

Archiving and group visibility is confusing. 
   1. you can't select an archived field to be set to active
   2. you **can** remove group visibility with the menu item, which expands archived items, select a field that is actually archived and thus have an active field that is archived. Further you can re enable group visibility and go to archived list and you will see an active archive item
   
The archive feature is not well defined anywhere, I don't think it should be treated like a group for the visibility menu button, and shouldn't be expanded like groups.

Otherwise the new dialog works really well and I like the addition of the neutral action archiving other non-active fields.

trife (changes requested):
- [ ] The active field indication is an issue that should be fixed even though the concept of active field will be removed in a later version.
- [x] I agree with @chaneylc that Archive should not be expanded when the other groups are. It should be treated as a special state.
- [ ] When in active mode (fields are selected), the following menu items should not be visible: group/ungroup, collapse/expand
- [ ] Toggle the group/ungroup icon depending on state. Use [this](https://pictogrammers.com/library/mdi/icon/ungroup/) icon as the toggle

bellerbrock:
Just tested this branch out and it looks great!

Maybe not in the scope of this first implementation, but would it make sense to map BrAPI trials to groups during BrAPI study imports?

trife:
@chaneylc I'm not seeing the same behavior with the selecting fields as you described it, but if you long-press on a group it will select all the children.

trife (approved):

trife:
@kamathprasad9 You will need to update the documentation as part of this PR. Please follow [these](https://github.com/PhenoApps/Field-Book/wiki/Documentation) guidelines for making those updates.

PR #1226, opened 2025-04-14
Fix trait bar resizing

### LINKED ISSUE #1200: Navigating back and forth between traits activity and Collect activity messes up sizing for the traits indicator
When navigating to Collect, the traits indicator box size is reset, but when navigating within Collect to other screens like traits and adjusting the traits enabled, the indicator box size is not reset and ends up with smaller/larger/mixed boxes. I think the reset size code needs called in onResume.

kamathprasad9:
### Description
Closes #1200 

### Change Type

### Release Note

```release-note
Trait indicator icons in Collect now resize as expected
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/adapters/TraitsStatusAdapter.kt
+++ b/app/src/main/java/com/fieldbook/tracker/adapters/TraitsStatusAdapter.kt
@@ -1,8 +1,5 @@
 package com.fieldbook.tracker.adapters
 
-import android.content.Context
-import android.util.DisplayMetrics
-import android.util.Log
 import android.util.TypedValue
 import android.view.LayoutInflater
 import android.view.View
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/fieldbook/tracker/adapters/TraitsStatusAdapter.kt
```diff
@@ -73,33 +76,40 @@ class TraitsStatusAdapter(private val traitBoxView: TraitBoxView) :
...
+            context.resources.getDimension(R.dimen.fb_trait_status_bar_icon_default_max_size_tablet_screen)
+        } else { // small screens
+            context.resources.getDimension(R.dimen.fb_trait_status_bar_icon_default_max_size)
+        }
+        TypedValue.applyDimension(
+            TypedValue.COMPLEX_UNIT_DIP,
+            defaultMaxSizeDp,
+            context.resources.displayMetrics
+        ).toInt() // in px
+    }
```
chaneylc, line 93: You should be able to provide alternative resources for device types instead of needing that dpi threshold check, which can lead to false tablet recognition:

https://developer.android.com/guide/topics/resources/providing-resources

chaneylc (changes requested):

PR #1229, opened 2025-04-25
Nix

### LINKED ISSUE #118: Nix Color Sensor Integration
https://www.nixsensor.com/nix-software-development-kit/

chaneylc:
### Description

closes #118

### Change Type

### Release Note

```release-note
Nix hardware sensors now supported as a new Spectral trait type
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


trife (changes requested):
- [x] Exporting in Table mode with only unique id and only active traits doesn't export correctly
- [ ] Probably should have a warning about only one of the values being exported in table mode if repeated measures exist
- [x] I'm not the biggest fan of the large floating progress bar.
- [x] Thoughts about overlaying the hex code within the color box?
- [x] When navigating between activities, the Nix disconnects and reconnects but the traitview doesn't update unless you switch plots.
- [x] Scroll to last collected value when navigating to plot with existing data

trife (changes requested):
@chaneylc It looks like when connecting to the sensor for the first time, network access is required. We should check for this and notify the user that they can't use the sensor if they're unable to connect their device to the Internet.

trife (approved):

PR #1234, opened 2025-04-29
Add vertical options, field grid preview

### LINKED ISSUE #254: Field creator changes
Creating a field should populate the position_coordinate_x_type, position_coordinate_y_type, position_coordinate_x, and position_coordinate_y fields. The x_type is row, the y_type is column, and the coordinates are their respective values.

kamathprasad9:
### Description
Closes #254 

### Change Type

### Release Note

```release-note

```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/dialogs/FieldCreatorDialogFragment.kt
+++ b/app/src/main/java/com/fieldbook/tracker/dialogs/FieldCreatorDialogFragment.kt
@@ -9,10 +9,18 @@ import android.util.TypedValue
 import android.view.View
 import android.widget.EditText
 import android.widget.ImageButton
-import android.widget.ImageView
-import android.widget.RadioButton
 import android.widget.TextView
 import android.widget.Toast
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.ComposeView
+import androidx.compose.ui.unit.dp
 import androidx.constraintlayout.widget.Group
 import androidx.fragment.app.DialogFragment
 import com.fieldbook.tracker.R
@@ -20,6 +28,11 @@ import com.fieldbook.tracker.activities.ThemedActivity
 import com.fieldbook.tracker.database.DataHelper
 import com.fieldbook.tracker.objects.FieldObject
 import com.fieldbook.tracker.objects.ImportFormat
+import com.fieldbook.tracker.utilities.FieldConfig
+import com.fieldbook.tracker.utilities.FieldPattern
+import com.fieldbook.tracker.utilities.FieldPlotCalculator
+import com.fieldbook.tracker.utilities.FieldStartCorner
+import com.fieldbook.tracker.views.FieldPreviewGrid
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.MainScope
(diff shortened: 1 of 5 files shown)


chaneylc (changes requested):
Creating a large field causes the UI to freeze. Once the dialog dismisses I still get random ANR dialogs until I close the app.

Probably need a progress bar / loading animation and use a coroutine for any intensive processing.

If an active field exists, and you create a new field it shows both as active.

Only insert row, col and plot as obs. unit attributes. The obs unit Id isn't needed here but I can deal with that in another PR. Remove position_coordinate_x/y and position_coordinate_x/y_type, these are BrAPI fields that help determine what the "row" "col" coordinates are, but since we are manually creating it within FieldBook these are non-essential. 

This seems to have added white space above and below the DataGrid activity.

I think some of the UI could be improved: centering the stepper, centering the stepper title text under the icon and styling it (maybe making it bold or a bit larger). Also The second "starting location" step is cool but I think it's easier to just have the old four buttons, especially for large fields, if we keep it with this grid-way then I think there should be some textual description on what the user has to do, the greyed-out boxes weren't immediately obvious for me, maybe a different pattern would help.

trife (changes requested):
- [x] Make this a full screen fragment
- [x] Grid on the starting point should not have numbers on the row/col
- [x] Shrink the grid down to fit the width of the window. You can add four fixed-size buttons near the corners that the user can use to choose their starting corner.
- [ ] I think the best is to split the walking order into two separate choices: serpentine/zigzag and then horizontal/vertical
- [ ] Try to shrink the preview grid down to fit the width of the screen. You can shrink the numbers too.
- [ ] Center the steps. Create a single centered label above the steps and rotate text through it instead of the text labels under each step.

I'm not seeing the white space above datagrid.

@chaneylc 
The existing method doesn't allow you to choose certain movement paths.
for pos_coord_x/y, do we not need this info to reliably create a map of the field without having to use other sources?

trife (changes requested):
- [x] Add next button to keyboard for field name edittext to go to rows edittext
- [ ] Add icons or images to the Linear/[Serpentine](https://pictogrammers.com/library/mdi/icon/arrow-u-left-top/) options (use the file i sent for linear)
- [ ] Add icons or images to the [Horizontal](https://pictogrammers.com/library/mdi/icon/table-row/)/[Vertical](https://pictogrammers.com/library/mdi/icon/table-column/) options
- [ ] Change text to "Walk up/down columns"
- [ ] Use [this](https://pictogrammers.com/library/mdi/icon/sign-direction/) icon for direction in stepper
- [ ] Use accent color for corner selection
- [ ] Change expand grid to FAB and use this [icon](https://pictogrammers.com/library/mdi/icon/arrow-expand-all/)
- [ ] Change Create to [checkmark](https://pictogrammers.com/library/mdi/icon/check-bold/) icon
- [ ] If on step > 1, ask user if they're sure they want to exit when trying to go back to fields page
- [x] Replace starting corner icon with [this](https://pictogrammers.com/library/mdi/icon/grid-large/)
- [x] Swap order of walking pattern and walking direction
- [ ] Add live field preview to each step- add numbers as they're known (e.g. corner, direction, order)

trife (changes requested):
- [x] Use the primary color for corners instead of accent
- [ ] Resizing the field preview between steps might cause confusion. Can we try keeping a 7x7 for all steps?
- [x] The current preview on walking pattern doesn't show until an option is selected. Can we keep the preview from the previous step visible?
- [x] Label all the plots at the walking pattern step
- [x] Indicate the corner in subsequent preview steps with color (try primary dark)
- [ ] Add a [collapse](https://pictogrammers.com/library/mdi/icon/arrow-collapse-all/) FAB to the full preview
- [ ] Remove text from full preview fragment
- [ ] PosCoordX/Y should follow the row/col numbers. In my testing, X is ordered high to low

chaneylc (changes requested):
Overall the code looks fine, usage of the stepper library with androidx fragments and viewmodels looks pretty solid. I would have liked to see more documentation on the fragments and how their viewmodel is used/connected.

I suggest these changes should be made at minimum:
- [ ] limit field size (or show a warning), currently its possible to crash the app by inputting very large fields
- [ ] entering row/col over max int shows "row/col must be positive"
- [ ] define string variables for argument keys
- [ ] view model validation errors need to be string resources 
- [ ] the "x/y_coordinate" and "x/y_coordinate_type" are unnecessary, unless I'm missing something these are planned for?

Minor thing:
I like the look of the final error message that shows the field size is too large, and was thinking adding this background cell to the row/col/field name would look nice.

trife:
@chaneylc The coordinate typing is inherited from brapi and is because type could be generic row/col but could also be lat/long. It's not fully necessary now but saves having to add it later when/if we start assigning spatial relationships based on geocoordinates.

trife (changes requested):
- [x] Resolve merge conflicts
- [ ] Collapsing preview after expanding doesn't shrink the preview

trife (approved):

PR #1236, opened 2025-05-04
Simplify traits attribute value access and storage operations

kamathprasad9:
### Change Type

### Release Note

```release-note

```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


app/src/main/java/com/fieldbook/tracker/database/dao/TraitAttributeValuesHelper.kt
```diff
@@ -0,0 +1,94 @@
...
+                    val attrId = value[ObservationVariableAttribute.FK] as? Int ?: return@forEach
+                    val attrName = ObservationVariableAttributeDao.getAttributeNameById(attrId).toString()
+                    val attrValue = value["observation_variable_attribute_value"] as? String ?: ""
+                    attributeValueMap[attrName] = attrValue
+                }
+                isLoaded = true
+                Log.d(TAG, "Loaded ${attributeValueMap.size} attributes for trait ID: $traitId")
+            } catch (e: Exception) {
+                Log.e(TAG, "Error loading attributes for trait ID: $traitId", e)
+            }
```
chaneylc, line 45: Let me know what you think, but I would prefer we simplify this further to one query or create a 'variable attribute details view' using a pivot query similar to the export queries, something like:

```
SELECT O.internal_id_observation_variable, O.observation_variable_name, 
MAX(CASE WHEN A.observation_variable_attribute_name = 'validValuesMax' THEN V.observation_variable_attribute_value ELSE NULL END) AS "validValuesMax",
MAX(CASE WHEN A.observation_variable_attribute_name = 'validValuesMin' THEN V.observation_variable_attribute_value ELSE NULL END) AS "validValuesMin"

... max case statements using your ALL static field ...

FROM observation_variables AS O
JOIN observation_variable_values AS V ON V.observation_variable_db_id = O.internal_id_observation_variable
JOIN observation_variable_attributes AS A ON A.internal_id_observation_variable_attribute = V.observation_variable_attribute_db_id
WHERE observation_variable_attribute_value IS NOT NULL AND observation_variable_attribute_value <> ""
GROUP BY O.internal_id_observation_variable
```

Then we can expand the TraitObject to initialize from this query which would look something like:

image

chaneylc (changes requested):
I like the improvements, but I think we can go a step further with the 'details' query I describe.

chaneylc:
It's also worth adding BrAPI unit attributes, look at the BrAPI usages of createFieldData when a study is saved in the service.

app/src/main/java/com/fieldbook/tracker/database/dao/TraitAttributeValuesHelper.kt
chaneylc: Re-reviewing this PR. I still think its worthwhile to simplify these queries further by taking this approach:
Going a step further, we could also use this query to create a developer view that allows easy-debugging/browsing of these attributes.

> Let me know what you think, but I would prefer we simplify this further to one query or create a 'variable attribute details view' using a pivot query similar to the export queries, something like:
> 
> ```
> SELECT O.internal_id_observation_variable, O.observation_variable_name, 
> MAX(CASE WHEN A.observation_variable_attribute_name = 'validValuesMax' THEN V.observation_variable_attribute_value ELSE NULL END) AS "validValuesMax",
> MAX(CASE WHEN A.observation_variable_attribute_name = 'validValuesMin' THEN V.observation_variable_attribute_value ELSE NULL END) AS "validValuesMin"
> 
> ... max case statements using your ALL static field ...
> 
> FROM observation_variables AS O
> JOIN observation_variable_values AS V ON V.observation_variable_db_id = O.internal_id_observation_variable
> JOIN observation_variable_attributes AS A ON A.internal_id_observation_variable_attribute = V.observation_variable_attribute_db_id
> WHERE observation_variable_attribute_value IS NOT NULL AND observation_variable_attribute_value <> ""
> GROUP BY O.internal_id_observation_variable
> ```
> 
> Then we can expand the TraitObject to initialize from this query which would look something like:
> 
> image

PR #1260, opened 2025-06-04
implemented end session request for oauth code token flow

chaneylc:
### Description

This implements the log-off request when clicking log-out in the brapi preferences for the oauth code token workflow. Previously, sessions would persist on the oauth server even after pressing log-out.

There's a slight lag when pressing the log-out button if code token is the chosen flow, this could be improved.

Implicit flow will be phased out at some point and we should eventually default to the code token flow.

### Change Type

### Release Note

```release-note
Improved workflow for logging out of BrAPI
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiAuthActivity.java
```diff
@@ -231,8 +199,10 @@ private void requestAuthorization(
                         redirectURI); // the redirect URI to which the auth response is sent
 
         if (!scope.trim().isEmpty()){
-            authRequestBuilder.setScope(scope);
+            authRequestBuilder.setScope(scope + " openid");
+            //authRequestBuilder.setScopes(scope, "openid");
         }
+        authRequestBuilder.setScopes("openid");
```
kamathprasad9, line 205: If the scope is not empty, the authRequestBuilder's scope will be overwritten - this probably needs to be an if-else block

kamathprasad9 (changes requested):
This looks great!! I think except for the possible minor fix, this works well.

trife (approved):

Issue #966, opened 2024-05-09
Add ability to edit BrAPI prefixes with user-defined values


trife:

trife:
I think this is closed with v7

PR #1271, opened 2025-06-20
Trait detail page

### LINKED ISSUE #18: [REQUEST]: Option to link traits to specific resource image files
This just came up in a field exercise where users had to scroll through many file names to load the correct resource image for a given trait. Many users wished the resource symbol could just automatically load the correct image. I'm thinking maybe this wouldn't be too hard to implement if the correct resource file could be identified in the trait definition?

### LINKED ISSUE #24: [REQUEST]: Automatic switch to next plot without confirmation on a per trait basis
In the software you have to "commit" (or confirm) when you are finished with the scoring of a plot, this is done with the "next >" button. As some of our breeders only score one trait per plot, this means doubling the number of clicks needed. 
Can you please insert a option to enable users to "jump to the next plot" after a user defined number of traits (or at least for a small and simple solution after one trait)?

Please contact me for further questions, and many thx for you and your great project!

### LINKED ISSUE #64: Add units to trait creation process

### LINKED ISSUE #135: [REQUEST] Consider changing numeric keypad to remove symbols for mathematical operations
The current numerical keypad includes signs for carrying out operations, e.g.+, *, etc. It seems like these should not be used to generate numerical data values (especially because they keypad does not seem to function like a calculator). Nevertheless, users may hit them by accident. 

Would it be possible to bring in a different numerical keypad that just has the 10 digits plus a decimal point if these other types of symbols should not be included in numerical data?

### LINKED ISSUE #447: Modify multiple measurements per trait to be on a per trait basis instead of global

### LINKED ISSUE #450: Option to view traits by name or synonym (BrAPI) or to give custom alias

### LINKED ISSUE #542: [BUG] Using obs. var names instead of synonym names
**Describe the bug**
This should just use the obs. var name. When importing traits from brapi.

### LINKED ISSUE #862: TraitEditorAcitivity updates
Similar changes to #148 - I'll work on a mockup

- [ ] Split import/export
- [ ] Use sort dialog from fields
- [ ] Include dialog with trait metadata 
- [ ] Trait metadata should include actions currently in menu

### LINKED ISSUE #896: [REQUEST]: Convert day of year setting into a setting in the date trait format, or it's own format
### Description

Suggested after a discussion with the BIMS Field Book training group. Some of their ontology traits should be collected with traditional date values, some should be collected with the Julian date (1-365). It would be difficult to impossible to manage the collection of both using the global setting.

### Alternatives

_No response_

### Additional Context

_No response_

### LINKED ISSUE #932: [BUG]: Modify date data storage to keep both date and day of year
### Description

Originally opened as TriticeaeToolbox/breedbase#6.

Date trait exports data sometime in Julian day of the year, sometimes as a date string. In both cases, they appear in Fieldbook's "Collect" view in the standard way (ie, "Mar 21"). 

image

### Expected behavior

_No response_

### Field Book version

_No response_

### LINKED ISSUE #933: [REQUEST]: Add the ability to limit or require number of decimal places that can be recorded for a numerical trait.
### Description

For a numerical trait, I want to be able to define the number of decimal places required for recording a value or limit the value to an integer only. 

### Alternatives

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1227: [BUG]: Calendar in date traits does not respect day of year setting
### Description

Saves full date instead

### Expected behavior

_No response_

### Field Book version

6.2.1

### LINKED ISSUE #1235: Add option to traits that asks if invalid value should be allowed
For outliers

### LINKED ISSUE #1279: Combine categorical and multicat traits with new option to allow multiple categories to be selected

### LINKED ISSUE #1336: Fix getAllFromAYear in ObservationDao to join tables

trife:
### Description

Closes #862
Closes #18 
Closes #450 
Closes #542 
Closes #896 
Closes #932 
Closes #135 
Closes #933 
Closes #1279 
Closes #447 
Closes #64 
Closes #24
Closes #1227 
Closes #1235 
Closes #1336 

### Change Type

### Release Note

```release-note
Traits list updated to include individual trait pages
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/utilities/export/SpectralFileProcessor.kt
+++ b/app/src/main/java/com/fieldbook/tracker/utilities/export/SpectralFileProcessor.kt
@@ -29,8 +29,8 @@ class SpectralFileProcessor @Inject constructor(private val database: DataHelper
             val decodedData = String(data)
             val values = decodedData.split(";")[0]
             return Result.success(when {
-                values.length == COLOR_DATA_SIZE -> fact.color
                 trait.format == Formats.GREEN_SEEKER.getDatabaseName() -> fact.color
+                values.length <= COLOR_DATA_SIZE -> fact.color
                 else -> uri
             })
         } else throw ProcessException.InvalidValue


trife (commented):
- [ ] Remove global repeated measures setting
- [ ] Currently name field/alias swapping is only available for brapi traits but should be an option for all traits
- [ ] Add per-trait options to the Options card
- [ ] Since Categorical value/label is only relevant for brapi traits, only display for brapi traits
- [ ] #542 
- [ ] #966 
- [ ] Picking dates from calendar doesn't automatically save (related to #1227 )
- [ ] #1235

trife:
@kamathprasad9 

- [ ] There's an extra toolbar in the per trait pages (probably due to the API updates)
- [ ] Change chip vertical spacing to match field details page
- [ ] Make the "Visible" chip toggleable (and without a background)
- [ ] Rename chip labels to: "Resource file", "Rename", "Outlier values", "Math symbols", "Multiple categories", "Automatic advance", "Format: nnn" (for date), "Transfer images"
- [ ] Only first word in other chip names should be capitalized
- [ ] Icons: [Outliers](https://pictogrammers.com/library/mdi/icon/chart-bell-curve/), [toolbar edit](https://pictogrammers.com/library/mdi/icon/cog-outline/), [rename](https://pictogrammers.com/library/mdi/icon/rename/), [auto switch](https://pictogrammers.com/library/mdi/icon/debug-step-over/), [copy image to field book (gopro)](https://pictogrammers.com/library/mdi/icon/progress-upload/)
- [ ] Thoughts about turning chips for boolean settings into icon-only changes? e.g. [on](https://pictogrammers.com/library/mdi/icon/eye/)/[off](https://pictogrammers.com/library/mdi/icon/eye-off/). Would require some custom icons.
- [ ] Fix the [disease rating trait icon](https://pictogrammers.com/library/mdi/icon/bug/) in list item
- [ ] I think the scrollview containing the list of traits needs to go to parent bottom instead of wrapping contents

chaneylc:
Still going over a lot of this but here are a few minor things I found:

Going back to trait editor after setting a trait alias doesn't show the new alias until the page is reopened.

Copied and imported traits show a blank text in their source.

It's possible to delete every category from a categorical trait, you can also create one with no categories (not sure if that was always possible)

Saving "Unit" option shows a toast that says "Edit", and I'm not sure how that "Unit" is currently being used.

Might be a larger issue from reused code, but every so often when entering the trait details page I see a large list of "A resource failed to call release or close." in logcat

app/src/main/java/com/fieldbook/tracker/activities/brapi/BrapiTraitActivity.java
```diff
@@ -258,9 +258,10 @@ public String saveTraits() {
             TraitObject trait = selectedTraits.get(i);
```
chaneylc: Encapsulate trait saving into a TraitSaver class so that we have a SSoT.

app/src/main/java/com/fieldbook/tracker/async/ExportJsonTraitTask.kt
```diff
@@ -0,0 +1,172 @@
...
+
+            if (!traitDir.exists()) {
+                return ExportResult.Error(context.getString(R.string.export_error_directory_not_found))
+            }
+
+            val exportDoc = traitDir.createFile("*/*", fileName)
+                ?: return ExportResult.Error(context.getString(R.string.export_error_file_creation))
+
+            val output = BaseDocumentTreeUtil.getFileOutputStream(context, R.string.dir_trait, fileName)
+                ?: return ExportResult.Error(context.getString(R.string.export_error_file_stream))
```
chaneylc: Potential memory leak here if output doesn't close, replace with "use" or close in finally.

PR #1311, opened 2025-08-20
adds new brapi traits to end of the trait list

### LINKED ISSUE #1252: [BUG]: BrAPI imported trait gets assigned position 0 in new UI
### Description

Instead of incrementing the last trait's position, 0 is assigned to all BrAPI imported traits in the new Import UI

### Expected behavior

_No response_

### Field Book version

_No response_

chaneylc:
fixes #1252

### Change Type

### Release Note

```release-note
New traits from BrAPI now appear at the end of the trait list
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/brapi/io/BrapiStudyImportActivity.kt
+++ b/app/src/main/java/com/fieldbook/tracker/activities/brapi/io/BrapiStudyImportActivity.kt
@@ -51,6 +51,7 @@ import org.brapi.v2.model.pheno.BrAPIPositionCoordinateTypeEnum
 import java.util.Locale
 import javax.inject.Inject
 import kotlin.collections.set
+import kotlin.math.max
 
 /**
  * receive study information including trial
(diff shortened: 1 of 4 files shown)


kamathprasad9:
This works well for brapi trait imports. But I just noticed that the issue still persists for traits imported while importing a brapi study

kamathprasad9 (changes requested):
This works well for brapi trait imports. But I just noticed that the issue still persists for traits imported while importing a brapi study

kamathprasad9 (changes requested):
Brapi study imported when the app has no traits saved previously results in position indexes starting from -1 for the corresponding traits. 

Moreover, for brapi studies we are doing `db.allTraitObjects.size - 1`, whereas for brapi trait import, we just have `database.allTraitObjects?.size`. In my testing, both these resulted in unexpected values sometimes. I suspect there could be a different problem.

@chaneylc I'm not sure if this could possibly be a new issue:
* delete any (except the last) trait from the traits list (eg. 11 traits)
* The position attribute for other traits which were below the deleted trait isn't updated. So the last trait with position 10 stays 10
* The consequent brapi import uses allTraitObjects.size, which will not be a clear reflection of what the max real position is

So either we update the real position while deleting traits, or we use the max value of position column in traits

chaneylc:
Thanks for delving into this. I forgot we have a query for this! I updated trait importer, study importer and the brapi sync to use this query to find the max position. The trait editor drag position code also started from 0, so updated that as well.

kamathprasad9 (approved):

PR #1316, opened 2025-08-25
Stop Watch Trait

### LINKED ISSUE #1077: [REQUEST]: Add timer/stopwatch trait
### Description

A timer/stopwatch trait would be quite useful for recording existing quality traits like fry cooking time in sweetpotato. And could facilitate the conversion of traits that are currently categorical (sweetpotato root attachment strength, cassava root peelability) into more quantitative measures like detachment time, peeling time, etc.

I'm thinking it'd be a stopwatch like in the android clock app, though only showing second-level precision. With parameters that can be set on creation, like start time, max time, etc.

### Alternatives

_No response_

### Additional Context

_No response_

chaneylc:
### Description

max time 99:59:59

database saved format example: 01:01:01 for one hour, one minute, and one second

features: reset, play/pause, save
resets time when deleted
matches current theme (iffy on the high contrast theme, need to test with palma)

fixes #1077

### Change Type

### Release Note

```release-note
New Stopwatch trait format
```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -314,7 +314,7 @@
     <string name="traits_format_spectral">Spectral</string>
     <string name="traits_format_nix">Nix</string>
     <string name="traits_format_inno_spectra">InnoSpectra</string>
-    <string name="traits_format_stop_watch">Stop Watch</string>
+    <string name="traits_format_stop_watch">Stopwatch</string>
 
     <!-- Trait Creation Dialog -->
     <string name="traits_toolbar_add_trait">Trait Format</string>


trife (changes requested):
- [x] Track/display milliseconds
- [ ] Bigger buttons with more space between them (left align, center align, right align)
- [ ] Smaller time circle

trife (changes requested):
- [ ] Change order of operations when timer is running and save button is pressed. Currently it saves the value and then pauses which leads to the displayed and saved values being out of sync. Modify to pause the timer and then save.

trife (changes requested):
I did some more testing with this and think it needs additional changes. Right now the timer is continues to run even when you switch between entries. My expectation is that a single timer would be tied to a single entry:

- [ ] Switching entries should reset the timer to zero
- [ ] If the timer is running, the user should be prompted before the entry switch happens

trife (approved):

PR #1326, opened 2025-09-04
Improvements on variable attribute creation

chaneylc:
improved code to easily allow adding new var attributes.

### Change Type

### Release Note

```release-note

```
___

### Release Type

- [ ] **`MAJOR`**
- [ ] **`MINOR`**


kamathprasad9 (approved):
This is great! Simplifies the process of adding attributes to traits (without needing a migrator every time). Thanks for working on this @chaneylc

kamathprasad9:
@chaneylc Just noticed this, but guess that can be handled when save_image attribute is actually used in a trait - assign the `saveImage` attribute to the value passed in `editTraits` method in `ObservationVariableDao`

PR #1392, opened 2026-02-02
inno_spectra

chaneylc:
### Description

Inno Spectra Nano Trait

Potential improvements:
saving coefficient / reference matrix after first load to speed up subsequent connections

### Change Type

- [X] **`ADDITION`** (non-breaking change that adds functionality)

### Release Note

```release-note
New Innospectra trait format
```


trife (changes requested):
- [x] It seems that it's possible to scan using the physical button but the data isn't transferred until the software scan button is pressed. I think this is counter to how Prospector operated where the physical button would also initiate and save a scan.
- [x] There are issues with the layout when a device is disconnected- padding/spacing is off and the config button is still displayed. These issues are resolved by going back to Collect.
- [ ] Is this implementation device specific or will it work for other devices from InnoSpectra? Regardless, nano is not used in any of InnoSpectra's branding material, so if it's device specific we should refer to it by the model name (InnoSpectra NIR-S-G1) and if it's manufacturer specific we should refer to it by the company name (InnoSpectra -- there are a few places where a space is included)
- [ ] Include device battery level and firmware version in the config dialog
- [ ] Trait scroller doesn't show that data has been saved after a successful scan
- [ ] This is one for conversation: should scan numbers be dynamic or static? From a user perspective, if I scan something 5 times and then delete scan 2, having the remaining scans immediately renumber might be confusing
- [ ] I won't be able to immediately test this, but Innospectras have a calibrated range of wavelengths that they collect, i.e. the headers are not equal between devices (though they are close). How would collecting data from two different models in the same experiment impact the data organization and export?

trife (approved):

PR #1395, opened 2026-02-05
BrAPI Germplasm Handling

### LINKED ISSUE #1242: [REQUEST]: Gracefully Handle 403 Response From Non-essential Endpoints
### Description

In DeltaBreed, our Experimental Collaborator role has read access to a specific set of experiments they are invited to collaborate on. They do not have access to germplasm endpoints (such as /search/germplasm) and they must not be able access pedigree information or germplasm list information.

Field Book makes calls to the `/brapi/v2/germplasm` endpoint, which returns a [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/403) response if the authenticated user is an Experimental Collaborator when using Field Book with DeltaBreed.

It seems like all of the essential features of Field Book could operate without the data from that endpoint, the `/brapi/v1/observationunits` endpoint already returns `germplasmName`, for example. If Field Book continued to function without results from `/brapi/v2/germplasm` (even if some features or data in the app are absent or limited), that would support our use case.

### Alternatives

We will test with returning empty responses instead of 403 Forbidden, it is unknown how Field Book will handle this (attempted) workaround.

### Additional Context

_No response_

chaneylc:
fixes #1242

### Description
fix handling errors when germplasm calls fail or are empty for old brapi import

### Change Type

### Release Note

```release-note

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- BrAPI improvements and optimizations (https://github.com/PhenoApps/Field-Book/pull/1396)
 
 ## [v7.0.1] - 2026-02-01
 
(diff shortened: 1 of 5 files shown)


jloux-brapi:
Howdy.  I work with Breeding-Insight (was formerly part of BrAPI as well) and was just looking at this code because it seems additional info JsonObject handling here is a bit wonky and was breaking the DeltaBreed integration with FieldBook.

This new code does seem to at least get past import failures when the additional info looks a certain way, but there may be better ways to handle it and add it to the list of attributes, depending on what FieldBook expects to do with the additionalInfo data.

If you expect to not really do much with it, then I think this solution is fine.  If you intend to create a record of it and maintain it, or export it, I think this solution might need to be updated.

PR #1412, opened 2026-02-16
migrate brapi token management to accountmanager

### LINKED ISSUE #205: Replace/Extend current authentication with Android AccountManager
Allow other apps to use the same account for brapi authentication without having to login each time

https://developer.android.com/reference/android/accounts/AccountManager

trife:
### Description

Closes #205

### Change Type

### Release Note

```release-note
BrAPI tokens are now managed in Android system settings
```


trife:
Other apps call:

```
val am = AccountManager.get(context)
val accounts = am.getAccountsByType("org.phenoapps.brapi")
am.getAuthToken(accounts[0], "access_token", null, activity, callback, null)
```

kamathprasad9 (commented):
This works great!

If the user tries to login using two different accounts for the same server, the most recent token is saved - and I expect the workflow to be this way.

In addition to the changes mentioned, the `BrapiAccountHelper` and `BrapiAuthenticator` can be migrated to phenolib in the future. But regardless, this is working as intended

chaneylc:
I like the new compose UI

- [x] settings flash before hiding when navigating to brapi settings page
- [x] error shows when checking brapi compatibility chip for some servers (I see that it does work for some other servers like cassavabase), this does not cause a crash
- [x] after auth from add account/OS settings it leads to a blank white screen in field book but the account is added
- [x] Manual Input / Scan config could be combined in the add account stepper (full screen compose activity would be nice here)
- [x] BaseBrapiAccountDialog* class hierarchy is a little odd, it should use a view model and a UI State object to persist UI across lifecycle changes. I think a cleaner approach would be to implement it as one stepper within the add account activity, with a view model.
- [x] not confident in the mixed handling of extras/prefs it seems extras are sent to brapi auth from add account but will be dropped on the oauth intent call since that intent doesn't populate the extras, which may lead to using current preferences instead of original sent extras
- [x] editing an account seems to actually add new accounts
- [x] suspicious usage of addAccountExplicitly in BrapiAccountHelper when there are duplicate account display names, not entirely certain of the side effects of this but could it overwrite an existing accounts user data that has the same display name?
- [x] double-check new permission

chaneylc (approved):

PR #1462, opened 2026-04-21
Debug/trait resourcefile paths

Stathmin:
### Description

Some trait files store an absolute `resourceFile` value (often a SAF `content://` URI). When a project is copied between devices or storage locations (e.g., exporting `trait/*.trt` together with `resources/*` for offline use without a BrAPI server), those absolute references can no longer resolve. In Collect this can make resource files fail to open when tapped.

This change improves portability by supporting a stable `resources/<relative-path>` form for `resourceFile` and resolving it against the user’s configured Resources directory, while preserving legacy URI values for backward compatibility.

### Change Type

### Release Note

```release-note
Improved portability of trait resource file paths when moving Field Book projects between devices or storage locations.```


chaneylc:
@Stathmin I updated the resource finding a bit, please test on your side. It looked like your implementation was adding an extra "resources/" prefix even if it already existed. I also updated the display name finding with content resolver.

trife:
@Stathmin Please test this updated branch to see if it behaves as expected.

PR #1474, opened 2026-05-06
Feature/hide field nav arrows

Stathmin:
### Description

Adds an Experimental setting that enables a new Collect bottom toolbar button to hide/show the field navigation cluster (left/right arrows + row/col labels/values). When the setting is disabled, Collect behavior is unchanged (navigation cluster always visible and no extra button).

This helps when working with listed unordered stored samples (e.g., stored envelopes with ears of wheat): you can focus on data entry using search by id without needing a simulated field setup, and reducing the chance of misclicked navigation arrows.

### Change Type

### Release Note

```release-note
Added an Experimental option to hide/show field navigation arrows in the Collect screen.


Stathmin:
The new added button in the bottom bar of Collect now makes inputs more usable when working with multiple InfoBars. Also if the field is simulated (e.g.. we don't need the spatial data on the samples, because we are phenotyping some samples mailed to us from other facilities), it's always beneficial to hide these arrows :)

Screenshot_20260507_081853_com_fieldbook_tracker_debug_CollectActivity
Screenshot_20260507_081848_com_fieldbook_tracker_debug_CollectActivity
Screenshot_20260507_080923_com_fieldbook_tracker_debug_PreferencesActivity

trife:
I like the idea but not the implementation. I'll work on something that fits the flow for [this](https://github.com/PhenoApps/Field-Book/issues/1490) issue.

PR #1483, opened 2026-05-27
BrAPI Bugs

### LINKED ISSUE #452: BrAPI Value/Label Fix
The data saved to the database is currently whatever is displayed to the user, but should always be the value.

Need to also add a two way conversion function to move between saved value and label (if that's what the user has selected)

### LINKED ISSUE #550: [BUG] Failure during BrAPI Export results in invalid state
**Describe the bug**

Scenario:
During export, if there is an error that occurs after the server has stored the observation, and before Field Book updates the "lastSyncTime", then this can result in an invalid state in Field Book, where Field Book thinks a given observation has not been uploaded to the server.  If a user tries to upload again, they will most likely have a failure because an observation already exists (but Field Book doesn't know about it).

**Expected behavior**
Field Book should verify that an observation has been stored at the server by querying the server directly

 - Model: any
 - OS: any
 - Field Book Version: 5.3 (applicable to any version with BrAPI support)

### LINKED ISSUE #1452: Sync Categorical format of Trait data broken FB to BMS, and BMS to FB
### Description

This issue appears from FB V7+

1)When sync (uploading) from FN to BMS categorical data is no correctly parse by BMS (BMS issue), Data loaded to BMS as non valid text string containing category value (category name) and label (category description). (needs attention)

2)When sync (downloading) from BMS to FB categorical data is no correctly parse by FB, Data loaded to FB as text representing category value (category name) .(needs attention)

See example, In red categorical data collected and sent (sync) from FB to BMS; In yellow categorical data capture in BMS and sent (sync) to FB App. The late case, it is not correct, it is interpreted by Field Book as a string of characters.

 
Image

BMS team findings:

Upon checking, this is due to the change in the display for categorical values in the app. On version 6.3.6, I didn’t encounter the issue since the categorical values are displayed and being passed as the value itself and not as a JSON string (see screenshot below, first 2 rows are synced from the old version and the last 3 are from the new ones - tried both 7.1.2 and 7.2.0). Do we update our implementation of POST /{crop}/brapi/v2/observations (the one being called on upload of observations from the app) to handle both scenarios to accommodate both old and new versions of the app? One possible solution is to check if it passes a JSON string for categorical values and parse. If not, the string will be treated as the value.

Image

Per the BrAPI spec, the value field is defined as a string type, not JSON.
For reference: https://app.swaggerhub.com/apis-docs/PlantBreedingAPI/BrAPI-Phenotyping/2.0#/Observations/post_observations:~:text=ObservationNewRequest,ObservationNewRequest

Please let me know if there is something we are missing or need to be implemented on BMS side.

Cheers
Julian

### Expected behavior

_No response_

### Field Book version

_No response_

### LINKED ISSUE #1453: Sync data from FB App v7+ to BMS v31+ broken for date format TRAIT variables
### Description

Steps to reproduce

1. Create a study in BMS, add a Trait variable with date format
2. In FB app, add the field (BMS studybook)
3. Collect date forrmat data
4. Sync (upload) from FB app to BMS
5. Open the study in BMS to accept the collected data

Expected result

### Expected behavior

collected date format data is pending to be accepted in BMS

Actual result

there is no date format data to be accepted in BMS study. There is no error message in FB, and those date format TRAIT datapoints remains on FB app as new observations:

https://github.com/user-attachments/assets/42805e9f-6d55-4748-953b-974c92a02da1

### Field Book version

7.2.0

### LINKED ISSUE #1475: Unable to switch to new/downloaded values in the data grid
### Description

Currently, when a new value is downloaded, it doesn't automatically become the "active" selection. This leads to a few inconsistencies:

- **Data Grid**: Instead of updating the existing value, it shows both the original and the new one side-by-side (causing the ... truncation).
Image

- **Collect Display**: It continues to show the older value, even though a newer one is available. 
Image

- **Manual Effort**: Users have to manually find and select the new value from the grid to get the views back in sync. However, this is where the issue occurs, since the new value cannot currently be selected by users.

### Suggested Fix
Perhaps we can trigger a state refresh after a successful sync to ensure the "Latest" timestamped value is always the one the user sees first?

### Expected behavior

Allow users to select the new/downloaded value from EBS.

To keep the data clean and the workflow fast, it would be great if the system could:

- Auto-select the most recently downloaded value by default.
- Update all views (Grid and Collect) simultaneously so the user always sees the most current information without extra clicks.

### Field Book version

7.2

chaneylc:
### Description

Multiple brapi fixes with syncing data including date format and categorical format.

Fixed display of values in data grid.

Added sync validation to sync UI. This fixes FB-BrAPI observation connection if FB never receives back a BrAPI ID. 
1. Added unique device install ID to help match observations that came from a device, this check will also use the FB internal obs. id. 

Instead of re-uploading, a new dialog will show asking to push the observation as new or resync FB so it has the BrAPI ID and sync timestamp, the sync timestamp comes from the server's obs. timestamp.

Closes #1453, closes #1475, closes #1452, closes #550, closes #452 

### Change Type

### Release Note

```release-note
Numerous BrAPI bug fixes and improvements
```


jloux-brapi (approved):
I don't see any issues, looks good.

trife (approved):

PR #1487, opened 2026-06-10
epic/271-trait-viewer

### LINKED ISSUE #271: TraitViewerActivity
Group/divide traits by field/study

- [ ] #862 
- [ ] Create TraitViewerActivity [TVA] (duplicate existing TraitEditorActivity [TEA])
- [ ] Add TVA link to per field page
- [ ] Remove visibility checkboxes and dragsort from TEA when accessed from configActivity
- [ ] Add trait in TVA opens brapi-like dialog with existing selectable traits (and ability to add new) to import to TEA
- [ ] TraitBox should open field-specific TVA
- [ ] Add preference to use either per field trait list or single trait list

chaneylc:
field dependent trait lists

closes #271 

### Change Type

### Release Note

```release-note

```

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/activities/CameraActivity.kt
+++ b/app/src/main/java/com/fieldbook/tracker/activities/CameraActivity.kt
@@ -811,7 +811,7 @@ class CameraActivity : ThemedActivity() {
 
                     // Determine whether to show crop region overlay for video based on saved crop coordinates for this trait
                     val showCropForVideo = try {
-                        val key = GeneralKeys.getCropCoordinatesKey(traitId?.toInt() ?: -1)
+                        val key = GeneralKeys.getCropCoordinatesKey(traitId)
                         val cropCoordinates = prefs.getString(key, "")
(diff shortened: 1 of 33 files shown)


trife (changes requested):
- [x] I somehow changed the trait order from the field details screen. Both traits screens have height first, but field details lists location first which is what actually loads in Collect
- [x] Remove the index number in the traits lists
- [x] When no traits exist, going back to details and forward reloads all traits visually but all the traits are still actually missing (can't go to Collect). 
- [ ] Trait details visibility settings are accessible/saved in the main list- syncing field-traits matches the current value.
- [ ] Add a per field option that keeps the sub-list in sync with the main list (presence/absence and visibility). This can be a toggle that lives in the field details header card.
- [ ] In the Traits card in field details, let's just list how many traits in the same summary text as the collect/export instead of the icons and expandable list
- [ ] Remove text format from trait listitems
- [ ] Need a way of indicating that the sub trait list is different from the main trait list- preferably in the header
- [ ] Trait picker search should match other search layouts (full pill round), add selected should match other areas (e.g. brapi import), and add selected shouldn't show until traits are selected

PR #1492, opened 2026-07-07
Add pollinator trait for timed pollinator counts

BrianSmart:
### Change Type

### Release Note

```release-note
New Pollinator trait to count categories for a set duration
```


trife:
@BrianSmart This is awesome! I'll review over the next few days and let you know if we have any questions/follow up.

BrianSmart:
Thanks @trife! Pushed a few small changes since your comment so you'll see a slightly newer version. Now has readable exports instead of raw JSON, translatable category names, and screen-reader labels on the buttons. Happy to adjust anything to match your style better, no rush.

Export looks like this but happy to change if you have any ideas (or if you would recommend keeping JSON):
Bumble Bee: 9 | Honey Bee: 6 | Large Dark Bee: 10 | Small Dark Bee: 7 | Small Green Bee: 3 | Non-Bee: 5 | 5:00

trife:
@BrianSmart I went through and made a few changes to better follow some of the patterns across the rest of the app. Let me know if there's anything I changed that breaks the intended workflow. The trait is now housed in the "Custom" trait category when creating a new one and a few backend changes may require you to make a new trait from scratch since I didn't bother with code to migrate settings.

BrianSmart:
Thanks @trife! No worries at all, that makes way more sense. Appreciate your time on this.

I tested the new pollinator trait and went through everything, and it looks great other than a small bug on my part. I noticed if you have two pollinator traits, the categories don't update when you switch to the next one (until play/start is tapped, and only on plots you haven't collected yet). Basically: open pollinator trait 1, see it has categories A and B, then scroll to pollinator trait 2 and it will still show A and B (despite actually having 1 and 2). Did a simple fix for that but feel free to alter/reject, as it doesn't actually impact data/results.

chaneylc:
Looks good, I added some minor changes.

@BrianSmart would you prefer the trait to save each time a count is updated? I like how simple the implementation is right now, but I could imagine a scenario where you count many objects in the five minutes and the phone dies or crashes or a user accidentally navigates to the next plot, which would currently erase that tracking.

We could add an automatic pause for these scenarios so tracking could be resumed.

BrianSmart:
Thanks @chaneylc! Appreciate your feedback. I agree that it would be frustrating to count for 4 out of 5 minutes at one plot and then accidentally close the app or navigate and have to restart that plot. I think your idea for an automatic pause while preserving the counts is a great idea. I'd probably lean toward saving each count as it happens too, just so closing the app doesn't lose them. Would you like me to take a shot at that, or add it whichever way you find appropriate? Thanks again!

chaneylc:
No worries, I can make those changes.

chaneylc:
For now I kept it simple and observations save for each class count

BrianSmart:
> For now I kept it simple and observations save for each class count

Looks great, I like your implementation! Thanks so much.

chaneylc (approved):

trife:
@BrianSmart This will be part of 7.3 which will be released (at least to GitHub and beta Google Play) as soon as #1501 is merged

PR #1519, opened 2026-08-12
fix pollinator counts saving to the wrong entry

BrianSmart:
### Description

Fixes two bugs in the pollinator trait reported from the field, both in v7.3.0. Selecting the trait on an entry could copy the previous entry's counts onto it, which stored `finished` and left every control disabled so it could only be recorded after deleting it, or write zeros over an entry that was already recorded. `onExit` runs on the layout being switched to rather than the one being left, and the layout keeps its counts until it is shown again, so it was saving counts belonging to a different entry. The layout now tracks which observation its counts came from and skips the save once the entry has changed, and the exit save looks at counts taken here rather than `hasData`. Saving moves back to the main thread because `updateObservation` resolves the entry when it runs, so a background write lands on whichever entry is current by then.

### Change Type

### Release Note

```release-note
Fixed several issues with the Pollinator trait
```

### DIFF
--- a/app/src/main/java/com/fieldbook/tracker/traits/PollinatorTraitLayout.kt
+++ b/app/src/main/java/com/fieldbook/tracker/traits/PollinatorTraitLayout.kt
@@ -45,11 +45,7 @@ import com.fieldbook.tracker.traits.formats.parameters.DEFAULT_DURATION_SECONDS
 import com.fieldbook.tracker.ui.theme.AppTheme
 import com.fieldbook.tracker.utilities.CategoryJsonUtil
 import com.fieldbook.tracker.utilities.JsonUtil
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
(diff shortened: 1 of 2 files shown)


BrianSmart:
Relevant messages from my coworker:
"I had some problems with Field Book - and I think I can replicate them! I am available in the afternoons to talk about it, otherwise I will try to describe below: 
1) Not allowing a new pollinator trait to be saved. I have a trait "Head Count". I count all heads first. Then when I want to take pollinator observations I click through all "Head Count" entries until I find one that meets the threshold to take pollinator entries. After recording Pollinator Trait, I switch back to Head Count and keep clicking through entries. The next entry I try to record Pollinator trait in is greyed out and I can't click record. I have to delete it to get it to record a new entry.

2) Last entry does not save. For pollinator traits, the last entry I record does not save. It says I saw zero bugs for zero minutes. I tried a couple things but could not fix this bug."

chaneylc (approved):

PR #1522, opened 2026-08-14
improve gopro connection management

### LINKED ISSUE #1521: Go Pro Issues
### Description

I am having some Gopro issues using Fieldbook version 7.2.3 on a Lenovo Tab One  with a Gopro Hero 11.  I am having several issues.

1. If you connect to the Gropro and then leave the app or switch trials it is hard to reconnect the gopro. I find I have to toggle the wifi and bluetooth on and off to connect again.
2. The Gopro is susceptible to disconnecting every couple of photos if Copy Image to Field Book setting is set,
3.  Problem 2  doesn't seem to occur if Copy Image to Field Book is unset; however, when I look at the images on the GoPro they are named as generic "GOPRO123.jpg" rather than with the observation unit and trait name.

### Expected behavior

_No response_

### Field Book version

_No response_

trife:
### Description

Two user-reported bugs, both rooted in the GoPro integration never owning or releasing the resources it acquired.

**Reconnect after leaving the app or switching trials** required toggling Wi-Fi and Bluetooth. `connectGatt`'s handle was discarded and closed without a preceding `disconnect()`, leaking a GATT client interface on every failed connect; `WifiHelper` overwrote its `NetworkCallback` without unregistering. Both leaks live in the framework and outlive the activity. The API now owns the GATT handle, `WifiHelper` keeps a single registered request with `onLost`/`onUnavailable` and a timeout, and teardown is wired into `onStop`, `switchField`, back-press and trait exit — none of which released anything before. Connection state is now an explicit enum instead of being inferred from an ExoPlayer playback flag that stayed `true` after the AP dropped.

**Disconnects every few photos with "Copy Image to Field Book" on** were link contention. One shutter press fired two media queries, busy pollers stacked up on every plot navigation, and the whole JPEG was read into memory while the preview stream was still running. Capture is now a single serialized pipeline that quiets the preview, keep-alive and poller, then streams the download to disk. OkHttp is capped at two concurrent requests with a separate download timeout.

Also fixed along the way: `FfmpegHelper` reused a closed UDP socket, so every keep-alive after the first stream restart failed silently; observations are now inserted after bytes land rather than before; and the copy-off path actually records the filename instead of writing an empty `.jpg`.

Removes dead `GoProController.kt` and the orphan `trait_go_pro.xml`.

Backgrounding the app now ends the session and requires the user to press Connect again on return.

## Testing todos

**Reconnect**
- [ ] Connect → Home → return → Connect. No radio toggling.
- [ ] Connect → switch trial → Connect.
- [ ] Connect → switch trait → back → Connect.
- [ ] Connect → back out of Collect → re-enter → Connect.
- [ ] **10 connect/disconnect cycles in a row** (cycle 10 should behave like cycle 1).
- [ ] Power the camera off mid-session → trait returns to a usable Connect button.

**Capture**
- [ ] 10 consecutive captures with Copy Image **on** (no disconnects, 10 non-zero JPEGs, exactly 10 observations).
- [ ] 10 consecutive captures with Copy Image **off** (filename-only observations, no stray 0-byte files).
- [ ] Photo taken with the camera's own shutter button still files against the current plot.
- [ ] Rapid double-tap the shutter (second press rejected cleanly, no duplicate).
- [ ] Power the camera off mid-download (no observation row, no orphan file).
- [ ] Navigate plots while connected (logcat shows the preview is not rebuilt per plot and only one busy poller runs.)

**Regression**
- [ ] Canon trait connect/capture/disconnect (shares `WifiHelper`; `requestNetwork` now times out instead of hanging).
- [ ] Photo and USB camera traits still save (shared `saveJpegToStorage` path changed).

Closes #1521

### Change Type

### Release Note

```release-note
GoPro connection improvements
```


chaneylc (changes requested):
Improvements look good. Camera does not reconnect after powering it off in the collect activity.

trife:
@chaneylc I think I got the reconnection issue (and a few others) fixed

chaneylc (approved):

Issue #1523, opened 2026-08-14
Brapi Authentication Error in 7.3.0


ch728:
### Description

This error is new for 7.3.0, didn't have it for 7.2.3. When I try to authorize the brapi connection to my Breedbase instance it says "Error Starting BrAPI Auth".

trife:
Are you trying to reauthenticate or set up an entirely new connection? If the former, you may want to try completely removing the server and readding it and I can see if we can do a better job of migration. A lot was changed for how tokens are stored but the actual auth process didn't have too many changes.

If this is for a new connection, then would it be possible to set up a test account with limited scope/permissions on your instance for testing/recreation?

ch728:
New connection. Yes, I can set up an account for you to use for testing. I will follow up in an email.

trife:
I'm hopeful that 7.3.1 will solve this or at least provide a better message for why the auth is failing. We can reopen the issue if it persists with that release.

PR #1524, opened 2026-08-14
brapi accountmanager fixes

trife:
### Description

- Per-app account types — each app now owns org.phenoapps.brapi.<applicationId> (declared via a per-variant resValue) instead of one shared type, since only the app registering a type, or one co-signed with it, can write its accounts. Also removed the READ_TOKEN permission declaration, which was blocking debug and release from coexisting.
- Config sharing over a ContentProvider — BrapiConfigProvider/BrapiConfigClient publish each app's own accounts' non-secret config (URL, name, version, OIDC, has_token) to granted sibling packages, because getUserData is signature-gated; every former direct read now goes through accountInfo().
- Token borrowing — borrowToken() fetches a shared account's token from its owning app via getAuthToken and mirrors it locally; shared cards no longer offer Authorize/Edit/Remove, and storeToken/addAccountConfig refuse to create a second account for a server a sibling already provides.

Test with coordinate debug, field book debug, and field book release builds

### Change Type

### Release Note

```release-note
BrAPI account manager improvements and optimizations
```

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -79,6 +79,12 @@ android {
 
         buildConfigField "String", "NIX_LICENSE", "\"${nixLicense.replaceAll("\"", "\\\\\"")}\""
 
+        // The BrAPI account type this app owns, read by res/xml/brapi_authenticator.xml. It has to
+        // stay equal to BrapiAccountConstants.accountTypeFor(applicationId), which is what the code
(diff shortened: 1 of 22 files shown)


trife:
Not sure if this will fix #1523 but it should help debug the issue with better auth failed messages

chaneylc (approved):

PR #1531, opened 2026-08-26
print-trait-improve-865

chaneylc:
### Description

rework print trait layout UI with compose
added label print module for designing and previewing labels from parsed subset of ZPL

### Change Type

### Release Note

```release-note

```


trife (changes requested):
- [ ] Installing on top of my current test app did not migrate the database correctly, leading to `no such table: label_templates (code 1 SQLITE_ERROR)` crashes
- [x] ~~Move "Create New" template to bottom of list; remove +~~ After playing around with this, I think the template chip should just open directly to the template editor instead of showing the list first
- [x] Template editor fragment has edge-to-edge issues with three button navigation
- [x] Remove the redundant hint from the "Default Value" text box (i.e. "Shown when no property is assigned")
- [x] I don't think the dpi/dots translation is going to be helpful- just call it text size and remove the hint below that box
- [x] I don't understand what block width does from the hint or units
- [x] In Label Settings, I don't think the PW/LL line is helpful
- [x] In the label editor, move the edit/delete FABs to the same row as the add FAB (justify the edit/delete FABs on the left side)
- [x] I'm guessing it has something to do with wrapping, but I could never get pedigree to display in the preview
- [x] Add text element prefixes/suffixes
- [x] Make the toolbar title a fixed "Label editor" instead of changing with the template name
- [x] The template spinner text should not update when the name input below it is updated
- [x] The toolbar save icon needs to be different (e.g. more related to export) to differentiate it from the bottom save button
- [x] Use an undo icon instead of the refresh icon for reset
- [x] Date should just be an additional element type instead of an attribute option- the custom attribute options in the list was briefly confusing. This could also be a sub option for text elements instead of a different element type if that makes more sense.
- [x] In that same vein, I'm not sure how default value is handled if it's also an attribute option. Is it treated as static text when chosen or used when there isn't a value for a selected parameter?
- [x] Not sure if "blank" is needed as a parameter
- [x] If no printer connected, change the print button to a connect button.
- [x] Add a disconnect button to the settings (if this is a method that currently exists)
- [x] Are the number of printed labels still being recorded? Nothing was added to the cvEt after printing.
- [x] Label fields dialog should be opened from a button in the label configuration dialog instead of being its own button
- [ ] The module should be named label-print or label-print-provider

trife (changes requested):
- [ ] In the Collect settings: merge the connect/disconnect icons into a single pill button whose label changes depending on current state
- [ ] Possible to add an option in the settings for a default printing device (similar to what we do for GPS) that will automatically try to connect?
- [ ] For sizes, what if we just had a slider below the label preview to live adjust? If that's not possible, all of the size edittexts should be numeric keypad instead of text
- [ ] Add warning when leaving unsaved templates
- [ ] Add "default template" setting and use when making new traits
- [ ] Label editor toolbar title needs bold
- [ ] Deleted default templates are recreated when navigating to label editor

